PR #28, opened 2020-09-14
Initial commit of service map stateful processor plugin

AustinTag:
*Issue #, if available:*

*Description of changes:*
Adding initial commit of stateful service map processor plugin


dependency-versions.gradle
```diff
@@ -11,5 +11,6 @@ ext.versionMap = [
         reflections            : '0.9.11',
         slf4j_api              : '1.7.30',
         slf4j_log4j12          : '1.7.30',
-        validation_api         : '2.0.1.Final'
+        validation_api         : '2.0.1.Final',
+        opentelemetry_proto    : '0.8.0'
```
kowshikn, line 15: nit: Use 0.6.0, so that we are all using the same version.

AustinTag: Whoops, Ill make that update now

kowshikn (approved):
This looks good for the first revision. Just one thought,

```
            //TODO: Look into faster way to move cursor up N elements
            for(long i=0; i<start; i++) {
                cursor.next();
            }
```
If we can not optimize this we should do the whole scan and find keyranges once rather than each processor's iterate doing it.

erosas (approved):

PR #33, opened 2020-09-16
ENH: expose retry status config and handle retry in ES sink

chenqi0805:
*Issue #, if available:*
#9 

*Description of changes:*
This PR

(1) Exposes retry status as a list in the RetryConfiguration
(2) Added RetryUtils and test cases to control retry backoff time.
(3) Add retry handling in ElasticsearchSink

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #9, opened 2020-09-09
[ES sink] Add retry utilities in ES sink plugin


chenqi0805:
* Ensure ES High-level client does exponential retries.
* Handle 4XX errors (incl partial bulk failures)
* Retry 5xx forever with exponential backoff.
* Expose a file config, to which 4XX errors will be dumped.

PR #42, opened 2020-09-21
ENH: handle retry

chenqi0805:
*Issue #, if available:*
#9 

*Description of changes:*
This PR handles retry as discussed in #9 and #33 .


situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BackOffUtils.java
```diff
@@ -0,0 +1,30 @@
+package com.amazon.situp.plugins.sink.elasticsearch;
+
+import org.elasticsearch.common.unit.TimeValue;
+
+import java.util.Iterator;
+
+public final class BackOffUtils {
```
kowshikn, line 7: can we not make the initialAttempt as a part of this Util?

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BulkRetryStrategy.java
```diff
@@ -0,0 +1,141 @@
...
+                (e instanceof ElasticsearchException &&
+                        !NON_RETRY_STATUS.contains(((ElasticsearchException) e).status().getStatus())));
+    }
+
+    private Tuple<BulkRequest, BulkResponse> retry(final BulkRequest request) throws Exception {
+        final BulkResponse response = requestFunction.apply(request);
+        return Tuple.tuple(request, response);
+    }
+
+    private void handleRetry(
```
kowshikn, line 69: 1. Not sure if we need a tuple? we already have reference of the request/response of each recursive call.
2. We can avoid repeated calling of backofUtils.hasNext() by letting the next recursive call doing the check.
3. What is the issue with the recursion depth?

Check the below code?

```
        //if we move the initialAttempt as part of backOffUtils, basically bacofffUtils can return next() without any waittime the first time.
        if (backOffUtils.hasNext()) {
            backOffUtils.next();
            final BulkRequest newRequest = createBulkRequestForRetry(request, response);
            try {
                final Bulkresponse newResponse = requestFunction.apply(newRequest);
                if (newResponse.hasFailures()) {
                    //First handle non retryable with DLQ
                    handleFailures(newRequest, newResponse);
                    if (canRetry(newResponse)) {
                        handleRetry(newRequest, newResponse, backOffUtils);
                    }
                }
            } catch (final Exception e) {
                if (canRetry(e)) {
                    handleRetry(bulkRequestForRetry, false, backOffUtils);
                } else {
                    handleFailures(newRequest, e);
                }
            }
            
        } else {
            Log.error("No next attempt, this shouldn't happen");
        }
        return;
```

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BulkRetryStrategy.java
```diff
@@ -0,0 +1,141 @@
...
+                } else {
+                    handleFailures(tuple.v1().requests(), tuple.v2().getItems());
+                }
+            }
+        }
+    }
+
+    private BulkRequest createBulkRequestForRetry(final BulkRequest request, final BulkResponse response) {
+        if (response == null) {
+            // retry due to Exception
```
kowshikn, line 104: or first attempt.

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -119,18 +121,13 @@ private long calcEstimatedSizeInBytes(final IndexRequest indexRequest) {
...
-      if (hasFailures) {
-        handleFailures(bulkRequest.requests(), bulkResponse.getItems());
-      }
-
-      return !hasFailures;
-    } catch (final IOException e) {
-      handleFailures(bulkRequest.requests(), e);
+      bulkRetryStrategy.execute(bulkRequest);
+    } catch (final InterruptedException e) {
+      LOG.error(e.getMessage(), e);
```
kowshikn, line 126: Its a best practice to have a message with error logs. 

``` LOG.error("some message",e);```

kowshikn (changes requested):

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BackOffUtils.java
chenqi0805: Checking...

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BulkRetryStrategy.java
chenqi0805: > 1. Not sure if we need a tuple? we already have reference of the request/response of each recursive call.

Thanks for the catching. Yeah, it is unnecessary now.

> 2. We can avoid repeated calling of backofUtils.hasNext() by letting the next recursive call doing the check.

Makes sense. that is redundant.

> 3. What is the issue with the recursion depth?

I am not sure if in Java there is a maximum recursion depth limit that if recursion depth exceeds it will throw an Exception.

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/ElasticsearchSink.java
chenqi0805: Here I do not have the exact message to describe about the interruption thrown by Thread.sleep, any suggestion?

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/BulkRetryStrategy.java
kowshikn: I would not worry about memory, because on exception we actually reuse the same bulkRequest object reference between recursive calls, which is the only realistic possible never ending recursion scenario.

kowshikn (approved):
LGTM.

AustinTag (commented):
Is there a reason that we don't want to just pull in an open source dependency for retries, such as retry4j?

https://github.com/elennick/retry4j

chenqi0805:
> Is there a reason that we don't want to just pull in an open source dependency for retries, such as retry4j?

@AustinTag Was not aware of that. Thanks for the call. I will reserve that for investigation and enhancement to serve our purpose. Currently the retry strategy is doing what we want.

AustinTag (approved):
Apart from potentially using an open source package for retries, this lgtm

PR #43, opened 2020-09-21
Add helper methods in PluginSettings

yadavcbala:
*Issue #, if available:*

*Description of changes:*
* Add helper method to PluginSettings for type specific attribute returns


kowshikn:
Awesome, could you add Boolean alone for now?

chenqi0805 (commented):
It will be great if we also have `getLongOrDefault`, which will facilitate bulk_size of ES sink.

chenqi0805 (approved):

kowshikn:
@chenqi0805 Are we not doing MB for BulkSize? (Adding Long is fine, I didn't propose long because I couldn't see any use case where we will need Long)

kowshikn (approved):
:yay:

PR #50, opened 2020-09-25
ENH: ES sink service map integration

chenqi0805:
*Issue #, if available:*

*Description of changes:*
This PR

1. Update service map index template
2. Extracts `hashId` as docId in service map data
3. Fixed index patterns for service map
4. Update README about service map config and custom data config
5. Added test coverage on service map

Note:

Instead of supporting two flags `trace.analytics.raw` and `trace.analytics.servicemap`, I used `index_type` that has 3 values: `raw`, `service_map`, `custom` which I think is more compact and simplifies the logic. I can return to the double flags if it leads to a better UX.

### DIFF
--- a/situp-plugins/elasticsearch/README.md
+++ b/situp-plugins/elasticsearch/README.md
@@ -70,7 +70,7 @@ Default value is false. If set to true, user also needs to set `trace_analytics_
 ### Custom data type
 
 If both `trace_analytics_raw` and `trace_analytics_service_map` are set to false, the ES sink will ingest custom data type, i.e.
-user needs to provide custom [`index_alias`](#index_alias).
+user needs to provide custom [index_alias](#index_alias).
 
 ### <a name="index_alias"></a>Index alias
 
--- a/situp-plugins/elasticsearch/README.md
+++ b/situp-plugins/elasticsearch/README.md
@@ -4,7 +4,9 @@ This is the SITUP Elasticsearch sink plugin, sending records to Elasticsearch cl
 
 ## Configuration
 
-The Elasticsearch sink should be configured as part of SITUP pipeline yaml file. An example configuration is as follows
+The Elasticsearch sink should be configured as part of SITUP pipeline yaml file according to the following use cases.
+
+### Raw span trace analytics
 
 ```$xslt
 pipeline:
@@ -13,11 +15,45 @@ pipeline:
     elasticsearch:
       hosts: ["http://localhost:9200"]
       trace_analytics_raw: true
-      trace_analytics_service_map: false
-      template_file: /your/local/template-file.json
       dlq_file: /your/local/dlq-file
       bulk_size: 4
-``` 
+```
+
+The elasticsearch sink will reserve `otel-v1-apm-span-*` as index pattern and `otel-v1-apm-span` as index alias for record ingestion.
+
+### Service map trace analytics
+
+```$xslt
+pipeline:
+  ...
+  sink:
+    elasticsearch:
+      hosts: ["http://localhost:9200"]
+      trace_analytics_service_map: true
+      dlq_file: /your/local/dlq-file
+      bulk_size: 4
+```
+
+The elasticsearch sink will reserve `otel-v1-apm-service-map` as index for record ingestion.
+
+### Custom data
+
+```$xslt
+pipeline:
+  ...
+  sink:
+    elasticsearch:
+      hosts: ["http://localhost:9200"]
+      index: "some-index"
+      template_file: /your/local/template-file.json
+      document_id_field: "someId"
+      dlq_file: /your/local/dlq-file
+      bulk_size: 4
+```
+
+User needs to provide custom index for record ingestion.
+
+## Parameters
 
 ### Hosts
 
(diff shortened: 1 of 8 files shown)


situp-plugins/elasticsearch/README.md
```diff
@@ -12,6 +12,9 @@ pipeline:
   sink:
     elasticsearch:
       hosts: ["http://localhost:9200"]
+      index_type: raw
+      index_alias: "apm-span"
+      template_file: /your/local/template-file.json
```
kowshikn, line 17: So, the users should not give these for trace-analytics feature. We should support configs like below,

```
trace_analytics_raw: true
or 
trace_analytics_service_map: true
```
Users can set only one of these configs to true. If they do, we should default the configs accordingly.

situp-plugins/elasticsearch/README.md
```diff
@@ -20,6 +23,203 @@ pipeline:
...
+        }
+      },
+      "traceGroupName": {
+        "ignore_above": 1024,
+        "type": "keyword"
+      }
+    }
+  }
+}
+```
```
kowshikn, line 219: nit: Can you link to the json file instead, this is not looking good for Readme.

chenqi0805: Sure. Agreed that should look much better.

situp-plugins/elasticsearch/README.md
```diff
@@ -12,6 +12,9 @@ pipeline:
   sink:
     elasticsearch:
       hosts: ["http://localhost:9200"]
+      trace_analytics_raw: true
+      trace_analytics_service_map: false
+      template_file: /your/local/template-file.json
```
kowshikn, line 17: Maybe a separate PR where we can actually write Readme per usecase. Example,

### Raw trace use case
```
sink:
    elasticsearch:
      hosts: ["http://localhost:9200"]
      trace_analytics_raw: true
      dlq_file: /your/local/dlq-file
      bulk_size: 4
```

### Service-map use case
```
sink:
    elasticsearch:
      hosts: ["http://localhost:9200"]
      trace_analytics_service_map: true
      dlq_file: /your/local/dlq-file
      bulk_size: 4
```
### General use case
```
sink:
    elasticsearch:
      hosts: ["http://localhost:9200"]
      index_alias: "name" # would prefer index instead of index_alias
      index_template: /path/to/index_template.json
      document_id: fieldName # this field will be extracted from the document and used as documentId
      dlq_file: /your/local/dlq-file
      bulk_size: 4
```

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -196,6 +202,16 @@ private String readTemplateURL(final URL templateURL) throws IOException {
     return parser.map();
   }
 
+  private String extractDocId(final Map<String, Object> source) {
+    if (indexType.equals(IndexConstants.RAW)) {
+      return (String) source.get("spanId");
+    } else if (indexType.equals(IndexConstants.SERVICE_MAP)) {
+      return (String) source.get("hashId");
+    } else {
+      return null;
```
kowshikn, line 211: nit: For generic we can extract from the user provided field.

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -152,7 +156,11 @@ public void stop() {
   private void createIndexTemplate() throws IOException {
     final String indexAlias = esSinkConfig.getIndexConfiguration().getIndexAlias();
     final PutIndexTemplateRequest putIndexTemplateRequest = new PutIndexTemplateRequest(indexAlias + "-index-template");
-    putIndexTemplateRequest.patterns(Collections.singletonList(indexAlias + "-*"));
+    if (indexType.equals(IndexConstants.RAW)) {
+      putIndexTemplateRequest.patterns(Collections.singletonList(indexAlias + "-*"));
+    } else {
+      putIndexTemplateRequest.patterns(Collections.singletonList(indexAlias));
+    }
```
kowshikn, line 163: nit: This code block and the if in Line#179 makes the sink code imperfect. I think we can do all the if else check on ```IndexConstants.Raw``` inside the indexConfiguration. You should consider this later.

kowshikn (approved):

situp-plugins/elasticsearch/src/main/java/com/amazon/situp/plugins/sink/elasticsearch/ElasticsearchSink.java
chenqi0805: So if I understand correctly, we can refactor index pattern and name into IndexConfig, right? Makes sense.

AustinTag (approved):
LGTM, I'll make sure the service map output reflects this configuration

PR #53, opened 2020-09-28
Adding hash id for relationships

AustinTag:
*Issue #, if available:*

*Description of changes:*

- Updating target and destination of ServiceMapRelationship to be Endpoint objects (with a resource and domain)
- Updating cache to check against actual object to avoid collisions with Integer hash code
- Adding hashId field to ServiceMapRelationship, which is an Md5 hash of all of the values for the object

Decided to use Md5 hash for document id because the 4 bytes for the integer hash code could lead to potential collisions when we get into the hundreds:

https://preshing.com/20110504/hash-collision-probabilities/

### DIFF
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapRelationship.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapRelationship.java
@@ -1,9 +1,18 @@
 package com.amazon.situp.plugins.processor;
 
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Objects;
+import com.sun.org.apache.xerces.internal.impl.dv.util.Base64;
 
 
 public class ServiceMapRelationship {
+
+    /**
+     * ThreadLocal object to generate hashes of relationships
+     */
+    private static ThreadLocal<MessageDigest> THREAD_LOCAL_MESSAGE_DIGEST = new ThreadLocal<>();
+
     /**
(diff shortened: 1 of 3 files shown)


situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapRelationship.java
```diff
@@ -110,20 +129,95 @@ public void setTraceGroupName(String traceGroupName) {
...
+        if(destination != null) {
+            result += destination.resource + "," + destination.domain;
+        }
+        return result;
+    }
+
+    private String md5Hash() {
+        if(THREAD_LOCAL_MESSAGE_DIGEST.get() == null) {
+            try {
+                THREAD_LOCAL_MESSAGE_DIGEST.set(MessageDigest.getInstance(MD5));
```
chenqi0805, line 172: Just curious. Is the usage of ThreadLocal to avoid thread competition on access to the "digest"(I mean the buffer) in MessageDigest?

chenqi0805 (approved):
LGTM. Non-blocking question.

kowshikn (approved):
LGTM

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapRelationship.java
AustinTag: Right. Its to avoid multiple threads using the "update" function at the same time and potentially merging two of the hashes. The other option was to lock this object, but its not an expensive object

PR #54, opened 2020-09-29
Builder class for OTel Trace Raw processor

sshivanii:
*Issue #, if available:*

*Description of changes:* This PR contains the Raw Trace Builder class built from resource, span and instrumentationLibrary protobuf objects.


situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+
+        public RawTrace build() {
+            return new RawTrace(this);
+        }
+
+    }
+
+    public static RawTrace buildFromProto(Span spans, InstrumentationLibrary instrumentationLibrary) {
```
kowshikn, line 92: buidlFromProto(Resouce resource, InstrumentationLibrary intrumentationLib, Span spans)

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,141 @@
...
+        private String instrumentationName;
+        private String instrumentationVersion;
+        private String serviceName;
+        private Map<String, String> resourceAttributes;
+
+        public RawTraceBuilder() {
+        }
+
+        public RawTraceBuilder setTraceID(String traceID) {
+            this.traceID = traceID;
```
yadavcbala, line 62: as suggested by @kowshikn go ahead and update the PR to include Preconditions.check for these

sshivanii:
> parentSpanId can be empty if this is a root span right ? are we handling them separately ?

Removing the check in parentSpanID in that case.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public static RawTrace buildFromProto(Resource resource, Span spans, InstrumentationLibrary instrumentationLibrary) {
+        final String SERVICE_NAME = "service.name";
+
+        return new RawTraceBuilder()
+                .setTraceID(Hex.encodeHexString(spans.getTraceId().toByteArray()))
+                .setSpanID(Hex.encodeHexString(spans.getSpanId().toByteArray()))
+                .setParentSpanID(Hex.encodeHexString(spans.getParentSpanId().toByteArray()))
+                .setName(spans.getName())
+                .setStartTime(convertStringNanosToISO8601(String.valueOf(spans.getStartTimeUnixNano())))
```
yadavcbala, line 136: can spans.getStartTimeUnixNano be 0 or null [think it cannot be null as it is long] ?
If its 0, ur conversion would return `1970-01-01T00:00:00` - is this fine ? or should we be doing something smart here ?

yadavcbala: @kowshikn its a nit but how do u think we should handle this - we can leave it and be 1970 or we can just put null and let sink (es) handle it

kowshikn: I think zero mapped to Unix epoch start tome is reasonable. If this is an issue we can address it later.

yadavcbala (approved):

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,156 @@
...
+        final long epochSeconds = nanos.divide(MILLIS_TO_NANOS.multiply(SEC_TO_MILLIS), RoundingMode.DOWN).longValue();
+        final int nanoAdj = nanos.remainder(MILLIS_TO_NANOS.multiply(SEC_TO_MILLIS)).intValue();
+        return Instant.ofEpochSecond(epochSeconds, nanoAdj).toString();
+    }
+
+
+    public static RawTrace buildFromProto(Resource resource, Span spans, InstrumentationLibrary instrumentationLibrary) {
+        final String SERVICE_NAME = "service.name";
+
+        Map<Object, Object> map = resource.getAttributesList().stream().collect(Collectors.toMap(keyVal -> keyVal.getKey(), keyVal -> keyVal.getValue()));
```
yadavcbala, line 134: Nit: `resource.getAttributesList().stream().collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue));`

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,156 @@
...
+        final Map<String, String> resourceAttributesMap = new HashMap<>();
+        for (Map.Entry<Object, Object> entry : map.entrySet()) {
+            if (entry.getValue() instanceof String) {
```
yadavcbala, line 137: do you expect all of the values to be instances of String here ?

what happens when there is a value like `new Boolean(true)` or `new Integer(23)` - we may end up ignoring those - are we good with it ?

sshivanii: Makes sense, let me modify this to convert those objects to String.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class RawTrace {
+
+    private static final BigDecimal MILLIS_TO_NANOS = new BigDecimal(1_000_000);
+    private static final BigDecimal SEC_TO_MILLIS = new BigDecimal(1_000);
+
+    private final String traceId;
```
kowshikn, line 23: add a field called `durationInNanos`

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,160 @@
...
+    private final String spanId;
+    private final String parentSpanId;
+    private final String name;
+    private final String startTime;
+    private final String endTime;
+    private final String instrumentationName;
+    private final String instrumentationVersion;
+    private final String serviceName;
+    private final Map<String, String> resourceAttributes;
```
kowshikn, line 32: Include span attributes.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
```diff
@@ -0,0 +1,160 @@
...
+        final String SERVICE_NAME = "service.name";
+
+        Map<Object, Object> map = resource.getAttributesList().stream().collect(Collectors.toMap(KeyValue::getKey, KeyValue::getValue));
+        final Map<String, String> resourceAttributesMap = new HashMap<>();
+        for (Map.Entry<Object, Object> entry : map.entrySet()) {
+            if (entry.getValue() instanceof String) {
+                resourceAttributesMap.put((String) entry.getKey(), (String) entry.getValue());
+            }
+            else {
+                resourceAttributesMap.put((String) entry.getKey(), String.valueOf(entry.getValue()));
```
kowshikn, line 142: We shouldn't do string.valueOf(), because it could be other [types](https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/common/v1/common.proto#L30-L36).

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
kowshikn: durationInNanos =  startimeinUnixNano - endTimeInUnixNano

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
yadavcbala: I think this is redundant now, probably think of doing it like below 

```
//if keyValue has null key, it will be mapped as null -> <whatever is the value>
        Map<String, String> resourceMap = resource.getAttributesList().stream()
                .collect(Collectors.toMap(
                        keyValue -> (String) keyValue.getKey(), 
                        keyValue-> String.valueOf(keyValue.getValue())));
```

or

```
       resource.getAttributesList().forEach( keyValue -> resourceMap.put((String) keyValue.getKey(), String.valueOf(keyValue.getValue())));
          
        });
```

or


```
        resource.getAttributesList().forEach( keyValue -> {
            if(keyValue.getKey() != null) {
                resourceMap.put((String) keyValue.getKey(), String.valueOf(keyValue.getValue()));
            }
        });
```
from the documentation of attributes, it does not look like there can be null keys - so u can ignore the null check of the key and just make the conversion. Also from the documentation it looks like the values are as below so we can safely assume all of the value objects have `toString()` implemented

```
// attributes is a collection of key/value pairs. The value can be a string,
  // an integer, a double or the Boolean values `true` or `false`. Note, global attributes
  // like server name can be set using the resource API. Examples of attributes:
  //
  //     "/http/user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
  //     "/http/server_latency": 300
  //     "abc.com/myattribute": true
  //     "abc.com/score": 10.239
  repeated opentelemetry.proto.common.v1.KeyValue attributes = 9;
```

kowshikn:
Include a Jackson Objectmapper and add a ```toJson``` method.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
yadavcbala: Seems like it can be `AnyValue` https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/common/v1/common.proto#L30-L36 - u can ignore the `String.valueOf()` completely because it can also be Array and KeyValueList

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/RawTrace.java
sshivanii: Will do

kowshikn (approved):

yadavcbala (approved):

PR #65, opened 2020-09-30
Update the service-map processor with config.

kowshikn:
* Added the plugin annotation and must have pluginsetting based constructor.
* Use processorsCreated instead of number of processors config.
* Default the LMDB path

### DIFF
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
@@ -1,5 +1,8 @@
 package com.amazon.situp.plugins.processor;
 
+import com.amazon.situp.model.PluginType;
+import com.amazon.situp.model.annotations.SitupPlugin;
+import com.amazon.situp.model.configuration.PluginSetting;
 import com.amazon.situp.model.processor.Processor;
 import com.amazon.situp.model.record.Record;
 import com.amazon.situp.plugins.processor.state.LmdbProcessorState;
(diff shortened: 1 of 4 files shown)


situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -39,28 +41,49 @@
...
+    }
+
+    ServiceMapStatefulProcessor(final long windowDurationMillis,
+                                       final File databasePath,
+                                       final File traceDatabasePath,
                                        final Clock clock) {
-        this.clock = clock;
+        ServiceMapStatefulProcessor.clock = clock;
         this.thisProcessorId = processorsCreated.getAndIncrement();
         if(isMasterInstance()) {
```
erosas, line 62: So if it is not the master instance the constructor does nothing? 
If thats the case it may be clearer to have an object hold all these variables and make that object an optional in this class. That would make it clearer that an instance of ServiceMapStatefulProcessor may or may not have a companion object with all these variables

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -138,6 +164,12 @@ public ServiceMapStatefulProcessor(final long windowDurationMillis, final File d
         }
     }
 
+    private static synchronized void initEdgeEvaluationLatch() {
+        if(edgeEvaluationLatch==null){
+            edgeEvaluationLatch = new CountDownLatch(processorsCreated.get());
```
AustinTag, line 169: Nice, this is cleaner/safer

AustinTag (approved):

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
kowshikn: MasterInstance() is true when the first object of ServiceMapStatefulProcessor is created. This is actually not perfect, but it will work for now. Ideally we would want to split the shared state to a separate class, but that would be a bigger change. So will do it later.

No Optional in the entire project :)

erosas (approved):

PR #72, opened 2020-10-01
Fix processors to work with our sample data

kowshikn:
*Description of changes:*
* The Raw processors was sending array list of json as the docs.
* The service-map processor was using a different key. 


sshivanii (approved):
LGTM

yadavcbala (approved):

PR #75, opened 2020-10-05
Add valid checks to spans before writing to state store.

kowshikn:
*Description of changes:*

1. Get service name and store the data in LMDB only if service name is available.
2. Check if span is valid and store only valid spans.

### DIFF
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
@@ -174,7 +174,7 @@ public class ServiceMapStatefulProcessor implements Processor<Record<ExportTrace
     private final BiFunction<byte[], ServiceMapStateData, Stream<ServiceMapRelationship>> previousWindowFunction = new BiFunction<byte[], ServiceMapStateData, Stream<ServiceMapRelationship>>() {
         @Override
         public Stream<ServiceMapRelationship> apply(byte[] s, ServiceMapStateData serviceMapStateData) {
-            return lookupLogic(serviceMapStateData, false);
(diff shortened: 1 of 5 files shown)


situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -178,17 +174,7 @@ private static synchronized void initEdgeEvaluationLatch() {
...
-            if (parentStateData != null && !parentStateData.serviceName.equals(serviceMapStateData.serviceName) && traceGroupName != null) {
-                return Arrays.asList(
-                        ServiceMapRelationship.newDestinationRelationship(parentStateData.serviceName, parentStateData.spanKind, serviceMapStateData.serviceName, serviceMapStateData.name, traceGroupName),
-                        //This extra edge is added for compatibility of the index for both stateless and stateful processors
-                        ServiceMapRelationship.newTargetRelationship(serviceMapStateData.serviceName, serviceMapStateData.spanKind, serviceMapStateData.serviceName, serviceMapStateData.name, traceGroupName)
-                ).stream();
-            } else {
-                return null;
-            }
+            return lookupLogic(serviceMapStateData, false);
```
erosas, line 177: a more descriptive name would be helpful here. Lookup logic for what>?

erosas (approved):
Changes look good, minor comment on naming of the lookup logic method.

I am a little worried about the cyclomatic complexity of this project. I understand that it's just a nature of the incoming data format and your methods encapsulating some of the logic really help. 

Let's add a task to integrate with https://pitest.org/ in order to check the quality of our test cases. 
Having a good mutation testing score would allow us to reduce complexity as it will discourage people from inlining a lot of logic as it will be much harder to test.

kowshikn:
@erosas Agreed, will look into pittest.

PR #76, opened 2020-10-07
Remove sleep and allow empty list to processors.

kowshikn:
*Description of changes:*
Small fix to make process worker work with Stateful processor.

### DIFF
--- a/situp-core/src/main/java/com/amazon/situp/pipeline/ProcessWorker.java
+++ b/situp-core/src/main/java/com/amazon/situp/pipeline/ProcessWorker.java
@@ -1,8 +1,8 @@
 package com.amazon.situp.pipeline;
 
-import com.amazon.situp.model.record.Record;
 import com.amazon.situp.model.buffer.Buffer;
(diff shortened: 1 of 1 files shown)


situp-core/src/main/java/com/amazon/situp/pipeline/ProcessWorker.java
```diff
@@ -38,31 +38,37 @@ public ProcessWorker(
...
+    }
+
+    /**
+     * TODO Implement this from Buffer [Probably AtomicBoolean], for now we will return true
      *
      * @return
      */
     private boolean isBufferEmpty() {
-        return isQueueEmpty;
+        return true;
```
yadavcbala, line 71: Why is it always true ?

kowshikn: > Implement this from Buffer [Probably AtomicBoolean], for now we will return true

Because we have to implement this in the buffer.

yadavcbala (approved):
Pending items will be addressed in next PRs

PR #78, opened 2020-10-07
Updating lmdbstate to use file passed in as db, and deleting db files…

AustinTag:
… on rotation

*Issue #, if available:*

*Description of changes:*

- Using the MDB_NOSUBDIR flag on the lmdb database creation. This allows you to pass in a File that corresponds to the database file to use (lock file is created as databaseFile + "-lock")
- Deleting database and lock files on window rotation
- Only passing in a single database path directory now, because all databases will be created as database files with unique names in that directory

### DIFF
--- a/situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
+++ b/situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
@@ -23,22 +23,24 @@ public class LmdbProcessorState<T> implements ProcessorState<byte[], T> {
     private final Dbi<ByteBuffer> db;
     private final Env<ByteBuffer> env;
     private final Class<T> clazz; //Needed for deserialization
+    private final File dbFile;
 
     /**
      * Constructor for LMDB processor state. See LMDB-Java for more info:
      * https://github.com/lmdbjava/lmdbjava
-     * @param dbPath The directory in which to store the LMDB data files
+     * @param dbFile The directory in which to store the LMDB data files
      * @param dbName Name of the database
      * @param clazz Class type for value storage
      */
-    public LmdbProcessorState(final File dbPath, final String dbName, final Class<T> clazz) {
+    public LmdbProcessorState(final File dbFile, final String dbName, final Class<T> clazz) {
         //TODO: These need to be configurable
         env = Env.create()
                 .setMapSize(10_485_760)
                 .setMaxDbs(1)
                 .setMaxReaders(10)
-                .open(dbPath, EnvFlags.MDB_NOTLS);
+                .open(dbFile, EnvFlags.MDB_NOTLS, EnvFlags.MDB_NOSUBDIR);
         db = env.openDbi(dbName, DbiFlags.MDB_CREATE);
+        this.dbFile = dbFile;
         this.clazz = clazz;
     }
 
@@ -113,6 +115,14 @@ public class LmdbProcessorState<T> implements ProcessorState<byte[], T> {
         try (final Txn<ByteBuffer> txn = env.txnWrite()) {
             db.drop(txn, true);
         }
+    }
+
+    public void delete() {
+        close();
+        final File lockFile = new File(dbFile.getPath() + "-lock");
+        //TODO: Add logging or metric here for failed deletions
+        dbFile.delete();
+        lockFile.delete();
     }
 
     @Override
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapProcessorConfig.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapProcessorConfig.java
@@ -3,6 +3,5 @@ package com.amazon.situp.plugins.processor;
 public class ServiceMapProcessorConfig {
     static final String WINDOW_DURATION = "window_duration";
     static final int DEFAULT_WINDOW_DURATION = 180;
-    static final String DEFAULT_TRACES_LMDB_PATH = "data/service-map/spans";
-    static final String DEFAULT_SPANS_LMDB_PATH = "data/service-map/traces";
+    static final String DEFAULT_LMDB_PATH = "data/service-map/";
 }
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
@@ -292,14 +292,14 @@ public class ServiceMapStatefulProcessor implements Processor<Record<ExportTrace
      * @return Next database name
      */
     private String getNewDbName() {
-        return "db-" + clock.millis();
+        return "/db-" + clock.millis();
     }
 
     /**
      * @return Next database name
      */
     private String getNewTraceDbName() {
-        return "trace-db-" + clock.millis();
+        return "/trace-db-" + clock.millis();
     }
 
     /**
--- a/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
+++ b/situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
@@ -38,8 +38,7 @@ public class ServiceMapStatefulProcessor implements Processor<Record<ExportTrace
     private volatile static LmdbProcessorState<String> currentTraceGroupWindow;
     //TODO: Consider keeping this state in lmdb
     private volatile static  HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
-    private static File databasePath;
-    private static File traceDatabasePath;
+    private static String dbPath;
     private static Clock clock;
 
     private final int thisProcessorId;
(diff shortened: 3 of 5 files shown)


situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -38,34 +38,30 @@
...
         if(isMasterInstance()) {
             previousTimestamp = ServiceMapStatefulProcessor.clock.millis();
             ServiceMapStatefulProcessor.windowDurationMillis = windowDurationMillis;
-            ServiceMapStatefulProcessor.databasePath = createPath(databasePath);
-            ServiceMapStatefulProcessor.traceDatabasePath = createPath(traceDatabasePath);
-            currentWindow = new LmdbProcessorState<>(databasePath, getNewDbName(), ServiceMapStateData.class);
-            previousWindow = new LmdbProcessorState<>(databasePath, getNewDbName() + "-prev", ServiceMapStateData.class);
-            currentTraceGroupWindow = new LmdbProcessorState<>(traceDatabasePath, getNewTraceDbName(), String.class);
-            previousTraceGroupWindow = new LmdbProcessorState<>(traceDatabasePath, getNewTraceDbName() + "-prev", String.class);
+            ServiceMapStatefulProcessor.dbPath = databasePath.getPath();
```
kowshikn, line 60: I guess this will not work, MDB_NOSUBDIR doesn't create the path to file. The reason testcases work is because Temporaryfolder.newfile() is valid. Add a test case with default config to confirm this works?

Note: The test case should create resource in the default config and delete it.

AustinTag: So im not sure why MDB_NOSUBDIR would need to create the path to the file. Without the flag it doesnt create the path either. Not really sure what the issue is here (I guess I can use that createPath function to be safe), but I can make the test case change.

kowshikn: The default path is `/data/service-ma/*`. If the directory doesn't exists the plugin crashes. 


`
Caused by: org.lmdbjava.LmdbNativeException$ConstantDerivedException: Platform constant error code: ENOENT No such file or directory (2)
	at org.lmdbjava.ResultCodeMapper.checkRc(ResultCodeMapper.java:114)
	at org.lmdbjava.Env$Builder.open(Env.java:495)
	at org.lmdbjava.Env$Builder.open(Env.java:512)
	at com.amazon.situp.plugins.processor.state.LmdbProcessorState.<init>(LmdbProcessorState.java:41)
	at com.amazon.situp.plugins.processor.ServiceMapStatefulProcessor.<init>(ServiceMapStatefulProcessor.java:61)
	at com.amazon.situp.plugins.processor.ServiceMapStatefulProcessor.<init>(ServiceMapStatefulProcessor.java:47)
	... 19 more
`

Fix,

`
 ServiceMapStatefulProcessor.dbPath = createPath(databasePath).getPath();
`

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapProcessorConfig.java
```diff
@@ -3,6 +3,5 @@
 public class ServiceMapProcessorConfig {
     static final String WINDOW_DURATION = "window_duration";
     static final int DEFAULT_WINDOW_DURATION = 180;
-    static final String DEFAULT_TRACES_LMDB_PATH = "data/service-map/spans";
-    static final String DEFAULT_SPANS_LMDB_PATH = "data/service-map/traces";
+    static final String DEFAULT_LMDB_PATH = "data/service-map/";
```
kowshikn, line 6:
```
~/open-source/simple-ingest-transformation-utility-pipeline/data master*
❯ ls 
service-map/                                 service-mapdb-1602165692694-prev             service-maptrace-db-1602165692695-lock
service-mapdb-1602165692085                  service-mapdb-1602165692694-prev-lock        service-maptrace-db-1602165692696-prev
service-mapdb-1602165692085-lock             service-maptrace-db-1602165692695            service-maptrace-db-1602165692696-prev-lock

```

Not sure if this the expected path you want to have?

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -38,34 +38,30 @@
...
-            currentWindow = new LmdbProcessorState<>(databasePath, getNewDbName(), ServiceMapStateData.class);
-            previousWindow = new LmdbProcessorState<>(databasePath, getNewDbName() + "-prev", ServiceMapStateData.class);
-            currentTraceGroupWindow = new LmdbProcessorState<>(traceDatabasePath, getNewTraceDbName(), String.class);
-            previousTraceGroupWindow = new LmdbProcessorState<>(traceDatabasePath, getNewTraceDbName() + "-prev", String.class);
+            ServiceMapStatefulProcessor.dbPath = createPath(databasePath);
+            currentWindow = new LmdbProcessorState<>(new File(dbPath.getPath() + getNewDbName()), "spansDb", ServiceMapStateData.class);
+            previousWindow = new LmdbProcessorState<>(new File(dbPath.getPath() + getNewDbName() + "-prev"), "spansDb", ServiceMapStateData.class);
+            currentTraceGroupWindow = new LmdbProcessorState<>(new File(dbPath.getPath() + getNewTraceDbName()), "traceDb", String.class);
+            previousTraceGroupWindow = new LmdbProcessorState<>(new File(dbPath.getPath() + getNewTraceDbName() + "-prev"), "traceDb", String.class);
```
kowshikn, line 64: 1. Why do we need `-prev`? When Situp starts for the first time there will be no previous right?
2. What is the logic for restart? If we are going with always start fresh policy? then wouldn't it mean we should have clean up job at the start up?

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapProcessorConfig.java
AustinTag: I see, looks like the "/" isnt added onto the directory path. Will fix this.

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
AustinTag: 1. It's a race condition that likely won't be hit, but also it's an issue that comes up in testing with a mock clock.
2. I think restart logic is not something that needs to be thought about and dealt with in another PR.

AustinTag: Oh, I see what you mean. We can leave it null (but then we have to have null checks on the window)

kowshikn: I think we shouldn't create empty files for the purpose of testing. I would prefer not to create previous window at all.

AustinTag: Well its not for the purpose of testing. Its an empty file thats deleted after the first window rotation so that we don't have to have null checks every time we access either previous window just for the startup window.

kowshikn: > Oh, I see what you mean. We can leave it null (but then we have to have null checks on the window)

I am okay with thanks checks.

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -280,14 +278,12 @@ private void resetWorkState() {
...
     private void rotateWindows() {
-        previousWindow.clear();
-        previousWindow.close();
-        previousTraceGroupWindow.clear();
-        previousTraceGroupWindow.close();
+        previousWindow.delete();
+        previousTraceGroupWindow.delete();
         previousWindow = currentWindow;
-        currentWindow = new LmdbProcessorState(databasePath, getNewDbName(), ServiceMapStateData.class);
+        currentWindow = new LmdbProcessorState(new File(dbPath.getPath() + getNewDbName()), "spansDb", ServiceMapStateData.class);
```
erosas, line 284: You could have just done String.join("/", part1, part2, par3)
In the future we should update the code to something like this. We can easily test the generation of each part and not have to worry about remembering which method will return with a /

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
erosas: Having an empty file will make the code consistent, if it only 1 place where this is checked then null check is fine but if it is several than it is less error prone (now and in the future) to always have the same flow.  Having an empty DB for a period of minutes is not a big deal as no significant resources would be consumed.

situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
```diff
@@ -115,6 +117,14 @@ public void clear() {
         }
     }
 
+    public void delete() {
+        close();
```
erosas, line 121: you can simplify this by deleting the entire directory. The parent directory contains both of these files.

erosas: oh wait, I just saw the EnvFlags.MDB_NOSUBDIR.

Does this mean that there will be a lock file with a unique name per DB?

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
AustinTag: So its more than just a single null check. We need to null check when we want to iterate over the previous window, when we want to look for parent span, when we want to find trace group, when we want to rotate the windows, etc. Its 5 or 6 null checks, and some which would happen quite a lot.

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
AustinTag: Yep, good point. I can make this change now

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
erosas: From  a complexity perspective I think we should keep this how you have it.
If this is a null check it means that that in order to have good test coverage we will need at most one new test per null check. Thats a lot more testing to avoid one empty DB file.

situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
AustinTag: talked offline

erosas (approved):

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
kowshikn: I'm okay to have empty file, but can we at least remove the `-prev` suffix because its not a true suffix, we attach it only  during the first run, after which there is no `-prev`. Maybe `-empty`. 

(Again this logic will change when we introduce on restart, I am bit worried that we are changing the algo for every PR rather try to have a full fledged idea)

AustinTag: Sure, I can change the suffix to "empty". I don't see how we are changing the algo every PR.

kowshikn (approved):

PR #80, opened 2020-10-08
Adding MapDB processor state and using it in Service Map Stateful Processor

AustinTag:
*Issue #, if available:*

*Description of changes:*

- Adding MapDB processor state plugin
- Using this plugin in the service map stateful processor
- Adding new, more generic, interface method for iterating over segments in processor state

### DIFF
--- a/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
+++ b/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
@@ -1,13 +1,11 @@
 package com.amazon.situp.plugins.processor.state;
 
 import com.amazon.situp.processor.state.ProcessorState;
+
+import java.io.IOException;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Random;
-import java.util.UUID;
+import java.util.*;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import org.junit.After;
--- a/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
+++ b/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
@@ -77,8 +77,7 @@ public abstract class ProcessorStateTest {
 
     @After
     public void teardown() {
-        processorState.clear();
-        processorState.close();
+        processorState.delete();
     }
 
     @Test
--- a/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
+++ b/situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
@@ -1,17 +1,15 @@
 package com.amazon.situp.plugins.processor.state;
 
 import com.amazon.situp.processor.state.ProcessorState;
-
-import java.io.IOException;
-import java.io.Serializable;
-import java.nio.charset.StandardCharsets;
-import java.util.*;
-import java.util.function.BiFunction;
-import java.util.stream.Collectors;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+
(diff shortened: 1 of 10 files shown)


situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
```diff
@@ -145,15 +145,30 @@ public long size() {
         }
     }
 
+    private long[] getRange(int segments, int index) {
```
erosas, line 148: you should return some sort of pair, I know Java is not the cleanest for this but the array is not intuitive

erosas (approved):

situp-plugins/mapdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/MapDbProcessorState.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Override
+    public void clear() {
+        map.clear();
+    }
+
+    @Override
+    public void close() {
+        map.close();
+    }
```
kowshikn, line 117: nit: From the first version of interface I have some odd feelings on this `clear` and `close`. I think we should just support one interface which is `delete`.  Delete works for any state implementation. Also, clear is expensive and close doesn't define cleaning up of resource. I agree mapdb support deleteOnClose, but other implementations won't.

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
```diff
@@ -225,10 +226,10 @@ private String getTraceGroupName(final byte[] traceId) {
      * Delete current state held in the processor
      */
     public void deleteState() {
```
kowshikn, line 228: This is not used, can we remove this?

situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
```diff
@@ -1,20 +1,16 @@
...
-import java.util.UUID;
-import java.util.function.BiFunction;
-import java.util.stream.Collectors;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.io.Serializable;
+import java.util.*;
```
kowshikn, line 10: nit: Avoid *.

situp-plugins/lmdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/LmdbProcessorState.java
kowshikn: ^ +1, maybe something like `KeyRange` object

AustinTag: Good point, ill make a KeyRange object. Pair requires javafx, which i'd rather not pull in if i dont have to

situp-plugins/mapdb-processor-state/src/main/java/com/amazon/situp/plugins/processor/state/MapDbProcessorState.java
AustinTag: That makes sense. I'll merge the interfaces into delete()

situp-plugins/service-map-stateful/src/main/java/com/amazon/situp/plugins/processor/ServiceMapStatefulProcessor.java
AustinTag: I was using it in the test class, but I can just manually clean the directory

situp-plugins/common/src/test/java/com/amazon/situp/plugins/processor/state/ProcessorStateTest.java
AustinTag: oops, yeah that was auto import. I will remove it

kowshikn (approved):

PR #91, opened 2020-10-15
Sample app which produces opentelemetry traces using opentelemetry libs

kowshikn:
*Description of changes:*

* excluded certain manifests as the docker images were crashing with missing main method. (@yadavcbala  please check this )
* Sample app that uses opentelemtery libraries


situp-core/build.gradle
```diff
@@ -30,5 +30,7 @@ jar {
     from {
         configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
     }
-
+    exclude 'META-INF/*.RSA'
+    exclude 'META-INF/*.SF'
+    exclude 'META-INF/*.DSA'
```
kowshikn, line 35: @yadavcbala please take a look at this.

yadavcbala: this will exclude the signature files from the jar - I don't think it will be a problem as we don't intend to sign the jar. But if possible, I would suggest to move the exclude to the jar task and see if that still solves the problem
```
jar {
    manifest {
        attributes('Implementation-Title': project.name,
                'Implementation-Version': project.version,
                'Main-Class': 'com.amazon.situp.SitupExecute')
    }
    from {
        configurations.runtimeClasspath.collect { it.isDirectory() ? it : zipTree(it) }
        exclude 'META-INF/*.RSA', 'META-INF/*.SF', 'META-INF/*.DSA'
    }
```

yadavcbala (approved):
All of these files are earlier reviewed in sample-app repo, Approving.

PR #92, opened 2020-10-16
Research exporting zipkin data from ES to Otel trace source

chenqi0805:
*Issue #, if available:*

*Description of changes:*
Reads data from specified zipkin index in ODFE to send to Otel trace source through grpc.


kowshikn (approved):
LGTM. We can address other fields like status message etc later.

PR #95, opened 2020-10-19
Add basic unit tests for Pipeline and Situp

yadavcbala:
*Issue #, if available:*

*Description of changes:*
* Add basic unit tests for Pipeline and Situp, will add more in subsequent PRs.
* Also update PluginSetting to avoid ClassCastException when getLong is called for int value attribute.

### DIFF
--- a/situp-core/src/main/resources/situp-default.yml
+++ b/situp-core/src/main/resources/situp-default.yml
@@ -2,8 +2,8 @@ otel-es-raw-pipeline:
   source:
     apm_trace_source:
   sink:
-    pipeline:
-      name: "otel-es-enrich-pipeline"
+    - pipeline:
+        name: "otel-es-enrich-pipeline"
   workers: 5
   delay: 3000
 otel-es-enrich-pipeline:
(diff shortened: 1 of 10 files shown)


situp-core/src/main/resources/situp-default.yml
```diff
@@ -2,15 +2,17 @@ otel-es-raw-pipeline:
...
   delay: 3000
 otel-es-enrich-pipeline:
   source:
     pipeline:
       name: "otel-es-raw-pipeline"
   sink:
-    elasticsearch:
-      hosts: ["https://search-sample-app-test-lqwynrnd2ikcuzfsrdilv4stbq.us-west-2.es.amazonaws.com"]
+    - elasticsearch:
+        hosts: ["https://search-sample-app-test-lqwynrnd2ikcuzfsrdilv4stbq.us-west-2.es.amazonaws.com"]
```
laneholloway, line 15: This really should not be in the default yaml.

laneholloway (changes requested):
Can we remove the default es cluster value and make it something obvious like `DEFAULT_CLUSTER_NOT_SET` so the user knows that it needs to be changed; or that they need to point to their own configuration file?

situp-core/src/main/resources/situp-default.yml
kowshikn: Why should situp have a default yaml? This doesn't make sense, we should probable avoid this and ask users to always provide the yml fils as an argument.

yadavcbala: I agree with you but would prefer it have that change in its separate PR.

laneholloway (approved):
Approving so your next commit can remove the default file completely.

kowshikn (approved):

PR #97, opened 2020-10-21
Example: zipkin-sleuth-webmvc-example

chenqi0805:
*Issue #, if available:*

*Description of changes:*
Adds zipkin monitored frontend and backend example into SITUP-Otel. sleuth-webmvc-example is similar to brave-example except that it uses spring-mvc framework.

Note: for now there is still `spanId` issue when SITUP dealing with zipkin data. But one can still find some raw trace and Servicemap data. This should be rerun after fixing the issue.


yadavcbala (approved):
LGTM

kowshikn (approved):

PR #104, opened 2020-10-27
Updating service map benchmarks to take parameters for various fields…

AustinTag:
…, and adding README

*Description of changes:*
Making the service map benchmark more generic to take in parameters. Also adding README.

### DIFF
--- a/situp-benchmarks/service-map-stateful-benchmarks/READEME.md
+++ b/situp-benchmarks/service-map-stateful-benchmarks/READEME.md
@@ -1,0 +1,24 @@
+# Service Map Stateful Benchmarks
+
+This package contains benchmarks for the service map stateful processor using JMH: https://openjdk.java.net/projects/code-tools/jmh/ . 
+
+Integration with gradle is done with the following gradle plugin for JMH: https://github.com/melix/jmh-gradle-plugin.
+
+The plugin creates a source set for the JMH benchmarks, and provides a few gradle tasks for running and building the benchmarks.
+
+## Running the tests via gradle task
+
+Tests can be run via the "jmh" gradle task provided by the plugin. The README for the plugin provides the various parameters that
+can be provided to the plugin. 
+
+## Running the tests via JAR
+
+To run the tests via JAR, you can build the benchmark jar using the gradle task "jmhJar". This jar is an executable jar 
+that runs the benchmark tests. Example command:
+
+```
+java -jar service-map-stateful-benchmarks-0.1-beta-jmh.jar -t 8 -r 600 -i 2 -p batchSize=100 -p windowDurationSeconds=180
+```
+
+The above command will run the benchmarks with 8 threads, for 600 seconds (10 minutes) per iteration, 2 iterations. It also
+sets the batchSize and windowDurationSeconds benchmark parameters. 
--- a/situp-benchmarks/service-map-stateful-benchmarks/READEME.md
+++ b/situp-benchmarks/service-map-stateful-benchmarks/READEME.md
@@ -17,8 +17,8 @@ To run the tests via JAR, you can build the benchmark jar using the gradle task 
 that runs the benchmark tests. Example command:
 
 ```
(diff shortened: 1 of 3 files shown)


situp-benchmarks/service-map-stateful-benchmarks/READEME.md
```diff
@@ -0,0 +1,24 @@
...
+This package contains benchmarks for the service map stateful processor using JMH: https://openjdk.java.net/projects/code-tools/jmh/ . 
+
+Integration with gradle is done with the following gradle plugin for JMH: https://github.com/melix/jmh-gradle-plugin.
+
+The plugin creates a source set for the JMH benchmarks, and provides a few gradle tasks for running and building the benchmarks.
+
+## Running the tests via gradle task
+
+Tests can be run via the "jmh" gradle task provided by the plugin. The README for the plugin provides the various parameters that
+can be provided to the plugin.
```
chenqi0805, line 12: Might better also add the example command with args here

situp-benchmarks/service-map-stateful-benchmarks/READEME.md
```diff
@@ -0,0 +1,24 @@
...
+Tests can be run via the "jmh" gradle task provided by the plugin. The README for the plugin provides the various parameters that
+can be provided to the plugin. 
+
+## Running the tests via JAR
+
+To run the tests via JAR, you can build the benchmark jar using the gradle task "jmhJar". This jar is an executable jar 
+that runs the benchmark tests. Example command:
+
+```
+java -jar service-map-stateful-benchmarks-0.1-beta-jmh.jar -t 8 -r 600 -i 2 -p batchSize=100 -p windowDurationSeconds=180
```
kowshikn, line 20: Do you have results/summary of the benchmark? I am more interested in knowing the **records/second** with 1, 4, 8, 16 and 32 threads with default window duration. This would be useful to share with users who wish tune and also give us the baseline numbers.

situp-benchmarks/service-map-stateful-benchmarks/READEME.md
AustinTag: so the gradle task actually doesn't take arguments. you can only edit those parameters via the actual gradle task setup and overriding them. I did find some guidance that we can map them to project properties, if we need to.

situp-benchmarks/service-map-stateful-benchmarks/READEME.md
AustinTag: Sure, I do have numbers, but not up to 32 threads. Is that something we want to live in the README?

kowshikn: you could add,

situp-benchmarks/service-map-stateful-benchmarks/results/summary.md

situp-benchmarks/service-map-stateful-benchmarks/READEME.md
chenqi0805: I see. I misunderstood that this will take the same args just as below. It will be nice if it does. But I am OK with it.

chenqi0805 (approved):

situp-benchmarks/service-map-stateful-benchmarks/results/summary.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Setup
+These tests were run on AWS EC2 r5 generation instances. For benchmarking with CPU counts of 2, 4, 8, and 16, 
+instance types of r5.large, r5.xlarge, r5.2xlarge, and r5.4xlarge were used respectively. 
+
+For the benchmarking, iterations were run for 10 minutes, with the default window duration of 3 minutes. The batch size
+used, in terms of Records, was 1024. Throughput is represented in terms of Records processed per second, rather than
+batches processed per second.
+
+## Results
```
kowshikn, line 11: The summary looks good

kowshikn (approved):

PR #109, opened 2020-10-28
Refactor OTelTraceRawProcessor

kowshikn:
*Description of changes:*

1. Refactor the OTelTraceProcessor by using internal model objects. 
1.1. Use HexString for traceId, spanId and ParentSpanId
1.2.  Span.Attributes, Resource.Attributes, InstrumentationLibrary and SpanStatus will be a Map<String,Object> as these fields will be directly used in kibana.
1.3. Replace "." with "@" in keys, this is again to make it compatible for Elasticsearch
1.4. Add fields durationInNanos, serviceName.
1.5. Use ISO8601 for timestamp
2. Use the proto version from the version map
3. Update e2e to reflect the changes to the processor.

Note: Null checks are not a valid in the opentelemetry proto as the default instance is attached, so dropped them.


situp-core/src/integrationTest/java/com/amazon/situp/integration/EndToEndIntegrationTest.java
```diff
@@ -96,15 +98,10 @@ private void refreshIndices(final RestHighLevelClient restHighLevelClient) throw
...
-        WebClient.of().execute(RequestHeaders.builder()
-                        .scheme(SessionProtocol.HTTP)
-                        .authority("127.0.0.1:21890")
-                        .method(HttpMethod.POST)
-                        .path("/opentelemetry.proto.collector.trace.v1.TraceService/Export")
-                        .contentType(MediaType.JSON_UTF_8)
-                        .build(),
-                HttpData.copyOf(JsonFormat.printer().print(request).getBytes())).aggregate().join();
+    private void sendExportTraceServiceRequestToSource(ExportTraceServiceRequest request) {
+        final TraceServiceGrpc.TraceServiceBlockingStub client = Clients.newClient(
```
chenqi0805, line 102: Looks like e2e failed here. Need to add `com.linecorp.armeria:armeria-grpc:1.0.0` into the dependencies of `integrationTest.gradle`

chenqi0805: already fixed.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/OTelTraceRawProcessor.java
```diff
@@ -150,13 +61,24 @@ private static String convertStringNanosToISO8601(final String stringNanos) {
...
+                    final String serviceName = OTelProtoHelper.getServiceName(rs.getResource()).orElse(null);
+                    final Map<String, Object> resourceAttributes = OTelProtoHelper.getResourceAttributes(rs.getResource());
+                    for (InstrumentationLibrarySpans is : rs.getInstrumentationLibrarySpansList()) {
+                        for (Span sp : is.getSpansList()) {
+                            try {
+                                finalRecords.add(new Record<>(new RawSpanBuilder()
+                                        .setFromSpan(sp, is.getInstrumentationLibrary(), serviceName, resourceAttributes)
+                                        .build().toJson()));
+                            } catch (Exception ex) {
+                                log.error("Unable to process invalid Span {}:", sp, ex);
```
yadavcbala, line 74: Question: JFI: Were we doing the same before as well - if the processor fails to process even a single span, we will fail ?

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/model/OTelProtoHelper.java
```diff
@@ -0,0 +1,145 @@
...
+                } catch (JsonProcessingException e) {
+                    throw new RuntimeException(e);
+                }
+            default:
+                throw new RuntimeException("Unknown case");
+        }
+    }
+
+    public static Map<String, Object> getSpanAttributes(final Span span) {
+        return span.getAttributesList().stream().collect(Collectors.toMap(i -> SPAN_ATTRIBUTES_REPLACE_DOT_WITH_AT.apply(i.getKey()), i -> convertAnyValue(i.getValue())));
```
yadavcbala, line 84: Not a blocker: although its not the same code but it feels like repetitive code, can we do anything ?

yadavcbala (approved):

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/OTelTraceRawProcessor.java
kowshikn: So the request object contains 1 or more resourceSpans and each resourceSpan has 1 or more spans. We will skip the failed and move to next in both case. There is no reason for retry as it will not work.

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/processor/oteltrace/model/OTelProtoHelper.java
kowshikn: I wanted it to be tested explicitly rather testing based on input object instance. Also keeping it simple.

chenqi0805 (approved):
LGTM

AustinTag (approved):

PR #114, opened 2020-11-02
Throw the exceptions encountered during the start and stop of source.

kowshikn:
*Description of changes:*

* Update the OpenTelemetryTraceSource to throw the exceptions. As per the pipeline control flow, pipeline will stop if the source encounters issue at start.

Note:

* No testcase for stop because Aremeria framework handles the STOP very graciously. 

### DIFF
--- a/situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/source/oteltracesource/OTelTraceSource.java
+++ b/situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/source/oteltracesource/OTelTraceSource.java
@@ -30,6 +30,9 @@ public class OTelTraceSource implements Source<Record<ExportTraceServiceRequest>
 
     @Override
     public void start(Buffer<Record<ExportTraceServiceRequest>> buffer) {
+        if (buffer == null) {
+            throw new IllegalStateException("Buffer provided is null");
+        }
         if (server == null) {
             final ServerBuilder sb = Server.builder();
             sb.service(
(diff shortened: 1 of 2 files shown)


situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/source/oteltracesource/OTelTraceSource.java
```diff
@@ -54,30 +57,36 @@ public void start(Buffer<Record<ExportTraceServiceRequest>> buffer) {
...
 
     @Override
     public void stop() {
         if (server != null) {
             try {
                 server.stop().get();
-                LOG.info("Stopped otel_trace_source.");
             } catch (ExecutionException ex) {
-                LOG.error("otel_trace_source failed to stop due to:", ex.getCause());
+                if (ex.getCause() != null && ex.getCause() instanceof RuntimeException) {
```
kowshikn, line 79: The reason for doing these checks is to present the correct exception as possible, avoid wrappers.

situp-plugins/apmtracesource/src/test/java/com/amazon/situp/plugins/source/oteltrace/OTelTraceSourceTest.java
```diff
@@ -141,4 +142,22 @@ void testHttpFullBytes() {
                     validateBuffer();
                 });
     }
+
+    @Test
+    public void testDoubleStart() {
+        Assertions.assertThrows(IllegalStateException.class, () -> SOURCE.start(BUFFER));
```
chenqi0805, line 148: Not sure why this throws IllegalStateException while BUFFER is not null?

chenqi0805 (approved):
Just minor ques. LGTM

situp-plugins/apmtracesource/src/test/java/com/amazon/situp/plugins/source/oteltrace/OTelTraceSourceTest.java
kowshikn: This is because, we are trying to start a server that is already started. So it throws IllegalStateExeception.

chenqi0805: Have we and should we explicitly capture this exception in the `start()`? b/c otherwise this seems to be testing armeria instead of our API.

erosas (approved):

Issue #113, opened 2020-11-02
Update the raw index template


kowshikn:
The below are fields we should manage in the index template,

Properties

- traceId  - keyword - 256
- spandId - keyword - 256
- parentSpanId -keyword - 256
- name - keyword - 1024
- kind - keyword - 128
- startTime -  date_nanos
- endTime - date_nanos
- status.code - int
- status.message - keyword
- serviceName - keyword
- durationInNanos - long

Dynaminc Template (path_match)

- resource.attributes.* - keywords
- attributres.* -keywords

Other Fields (if possible)

- events - should be an array of nested objects
- links - should be an array of nested objects.

Note:

* With the current template, ```status.message``` is not friendly for kibana. (dicussed this on call)

PR #116, opened 2020-11-02
ENH: update raw span index template

chenqi0805:
*Issue #, if available:*
#113 

*Description of changes:*

### DIFF
--- a/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
+++ b/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
@@ -1,6 +1,6 @@
 {
   "mappings": {
-    "date_detection": true,
+    "date_detection": false,
     "dynamic_templates": [
       {
         "resource_attributes_map": {
--- a/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
+++ b/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
@@ -1,6 +1,6 @@
 {
   "mappings": {
-    "date_detection": false,
+    "date_detection": true,
     "dynamic_templates": [
       {
         "resource_attributes_map": {
@@ -66,7 +66,12 @@
         "type": "long"
       },
       "events": {
-        "type": "nested"
+        "type": "nested",
+        "properties": {
+          "time": {
+            "type": "date_nanos"
+          }
+        }
       },
       "links": {
         "type": "nested"
--- a/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
+++ b/situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
@@ -3,12 +3,19 @@
     "date_detection": false,
     "dynamic_templates": [
       {
-        "strings_as_keyword": {
+        "resource_attributes_map": {
           "mapping": {
-            "ignore_above": 1024,
-            "type": "keyword"
+            "type":"keyword"
           },
-          "match_mapping_type": "string"
+          "path_match":"resource.attributes.*"
+        }
+      },
+      {
+        "attributes_map": {
+          "mapping": {
+            "type":"keyword"
+          },
+          "path_match":"attributes.*"
         }
       }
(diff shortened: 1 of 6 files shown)


situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
```diff
@@ -38,9 +49,27 @@
       "endTime": {
         "type": "date_nanos"
       },
-      "resource.attributes.service.name": {
-        "ignore_above": 1024,
+      "status": {
```
kowshikn, line 52: Cool, this works. I tried running `status.code>0` in the KQL and this works.

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
```diff
@@ -38,9 +49,27 @@
...
         "type": "keyword"
+      },
+      "durationInNanos": {
+        "type": "long"
+      },
+      "events": {
+        "type": "nested"
+      },
+      "links": {
+        "type": "nested"
```
kowshikn, line 72: Just asking for my clarification, this definition will fail if index request is anyone of the following,
1. events: "some string"
2. events: { "atrributes": { "key":"value"}}

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
```diff
@@ -3,12 +3,19 @@
     "date_detection": false,
```
kowshikn, line 3: events is actually array of objects which contains a field called `time`. Now should we enable this flag or define the mapping events with `time` as `date_time_nanos`

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
chenqi0805: Yeah. If the above 2 coexists, define `events` as `nested` will lead to error. I do not know if any field type will work for the above scenario.

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
chenqi0805: Agreed.
```
{
  "time": "2020-11-03T00:01:19.277324Z",
  "name": "",
  "attributes": {
    "method": "GET",
    "level": "info",
    "event": "HTTP request received",
    "url": "/customer?customer=392"
  },
  "droppedAttributesCount": 0
}
```

chenqi0805: Maybe also enable the flag as there might be other fields fit into datetime

chenqi0805: Fixed

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
kowshikn: So the above 2 examples are both invalid, is there way to force and allow only,
`events: [ { time : ".....",  attributes : { key : "value" } } ]`

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
kowshikn: I think we shouldn't enable the flag we are explicitly mapping them. I am sure date_detection is an unwanted cost. So lets disable detection as we will going forward explicitly map date_time.

situp-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
chenqi0805: Let me check path_match for nested

kowshikn (approved):
Lets keep events and links as nested for now, I think if we have to add mappings, lets do it in the next iteration whenever we use those fields.

yadavcbala (approved):

PR #132, opened 2020-11-06
ENH: refactor flask services

chenqi0805:
*Issue #, if available:*

*Description of changes:*
- Build all flask services in a single docker image
- Run the above as a single container


chenqi0805:
Note: since we run multiple python processes in the background inside a single container now, I am looking into the way that trigger docker container to restart if one process goes wrong. But this should not be blocking for this PR.

kowshikn (approved):

Issue #151, opened 2020-11-13
Upgrade to latest OpenTelemetry Span Status spec.


kowshikn:
1. Upgrade SITUP to use the latest [version](https://github.com/opendistro-for-elasticsearch/simple-ingest-transformation-utility-pipeline/blob/da3c457817765f05fffaf6e6d10a7a6b9ff8af82/dependency-versions.gradle#L15) of OpenTelemetry Proto which has the new Span Spec. 

2. Upgrade the Sample apps to OpenTelemetryv0.14.0 which support the new Span Spec. 

3. Upgrade the sample also to latest OpenTelemetry python versions. 

4. After all this is done, perform an integration sanity check and ensure all sample apps provides only status code 0, 1 and 2. 

Note: We will do this in a [branch](https://github.com/opendistro-for-elasticsearch/simple-ingest-transformation-utility-pipeline/tree/otel-collectorv0.14.0) because this will break kibana dashboard as our dashboard uses ">0" as error code.

chenqi0805:
Kibana Dashboard after upgrade:

![Screen Shot 2020-11-13 at 2 13 57 PM](https://user-images.githubusercontent.com/19492223/99120391-66963280-25c0-11eb-8928-2fc87cb51ff0.png)

![Screen Shot 2020-11-13 at 2 13 37 PM](https://user-images.githubusercontent.com/19492223/99120341-58481680-25c0-11eb-9fb7-81b53eaec18d.png)

PR #153, opened 2020-11-13
Upgrade proto to 0.10.0

chenqi0805:
*Issue #, if available:*
#151

*Description of changes:*
- Upgraded proto dependency to v.10.0 and fixes the relevant test cases

Note: Will do an upgrade on sample app and do sanity check


chenqi0805:
Not correctly directed

Issue #152, opened 2020-11-13
Implement standard grpc health check


wrijeff:
Our current health check is based of HTTP1.1 as the ALB preview was did not support gRPC. ALB now supports gRPC health check. Implement HealthGrpc compliant with  https://github.com/grpc/grpc/blob/master/doc/health-checking.md and add it to the server (if health checking is enabled).

kowshikn:
As discussed offline, we should add flag for ProtoReflection so that users can use grpcurl and test the source.

wrijeff:
> As discussed offline, we should add flag for ProtoReflection so that users can use grpcurl and test the source.

Created https://github.com/opendistro-for-elasticsearch/simple-ingest-transformation-utility-pipeline/issues/160 for this.

PR #155, opened 2020-11-13
Implemented standard grpc health check.

wrijeff:
*Issue #, if available:* #152

*Description of changes:*
Implementing gRPC health check per [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md).

```
[ec2-user@ip-xxx-xx-xx-xxx ~]$ ./grpcurl -insecure  <address>:21890 grpc.health.v1.Health/Check
{
  "status": "SERVING"
}
```

### DIFF
--- a/situp-plugins/apmtracesource/build.gradle
+++ b/situp-plugins/apmtracesource/build.gradle
@@ -7,6 +7,7 @@ dependencies {
     compile project(':situp-api')
     compile project(':situp-plugins:common')
     compile 'commons-codec:commons-codec:1.15'
+    testCompile "io.grpc:grpc-testing:${versionMap.grpc}"
     implementation "io.opentelemetry:opentelemetry-proto:${versionMap.opentelemetry_proto}"
     implementation 'com.google.protobuf:protobuf-java-util:3.13.0'
     implementation "com.linecorp.armeria:armeria:1.0.0"
--- a/situp-plugins/apmtracesource/build.gradle
+++ b/situp-plugins/apmtracesource/build.gradle
@@ -7,7 +7,6 @@ dependencies {
     compile project(':situp-api')
     compile project(':situp-plugins:common')
     compile 'commons-codec:commons-codec:1.15'
-    testCompile "io.grpc:grpc-testing:${versionMap.grpc}"
     implementation "io.opentelemetry:opentelemetry-proto:${versionMap.opentelemetry_proto}"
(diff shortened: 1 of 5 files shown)


situp-plugins/apmtracesource/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     compile project(':situp-api')
     compile project(':situp-plugins:common')
     compile 'commons-codec:commons-codec:1.15'
+    testCompile "io.grpc:grpc-testing:${versionMap.grpc}"
```
kowshikn, line 10: Do we need this? If its only for the @Rule I think we can avoid this because we do not have direct dependency on grpc and this is introducing it. I would like to have dependency only on armeria.

wrijeff: Removed dependency. It's a nice to have but I wasn't able to find a way to keep Armeria's version of gRPC in sync with whichever version of grpc-testing we were pulling in.

kowshikn (approved):

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/health/HealthGrpcService.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class HealthGrpcService extends HealthGrpc.HealthImplBase {
+
+    @Override
+    public void check(final HealthCheckRequest request, final StreamObserver<HealthCheckResponse> responseObserver) {
+        if (isHealthy()) {
+            responseObserver.onNext(
+                    HealthCheckResponse.newBuilder().setStatus(HealthCheckResponse.ServingStatus.SERVING).build());
+        } else {
+            responseObserver.onNext(
```
chenqi0805, line 16: QUES: how is this block reached if isHealthy is always true? Is there something in the TODO?

chenqi0805 (approved):
Just a question for clarification. Otherwise LGTM!

situp-plugins/apmtracesource/src/main/java/com/amazon/situp/plugins/health/HealthGrpcService.java
chenqi0805: Yeah. Just for visibility to the whole team. I am OK with either way.

PR #166, opened 2020-11-18
Update to OpenTelemetryv0.14.0

kowshikn:
*Issue #151 *

*Description of changes:*

* Merging the upgrade to latest OpenTelemetry Span spec. Kibana has pushed the [change](https://github.com/opendistro-for-elasticsearch/trace-analytics/pull/3) and will be available part of ODFE Kibana 7.10.0 release scheduled on 12/10.
* In the sample use the jar from kibana team which uses ```status.code==2```


kowshikn:
Analytics service is not publishing trace data. Will investigate this one.

yadavcbala (approved):

examples/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM gradle:jdk14 AS builder
 COPY . /home/gradle/src
 WORKDIR /home/gradle/src/situp-core
-RUN gradle jar --daemon
+RUN gradle cleanjar --daemon
```
chenqi0805, line 4: This does not work on my laptop.

kowshikn: My bad, this was a typo. Fixing in the next commit. `RUN gradle clean jar --daemon`

kowshikn:
> Analytics service is not publishing trace data. Will investigate this one.

`[opentelemetry.auto.trace 2020-11-18 19:01:36:800 +0000] [grpc-default-executor-0] WARN io.opentelemetry.exporter.otlp.OtlpGrpcSpanExporter - Failed to export spans. Error message: UNAVAILABLE: io exception
 `
Looks like Analytics-service is having issues after version upgrade.

chenqi0805:
FTR offline discussion. There is manual introduction of random error:
```
@PostMapping("/logs")
    public String save(@RequestBody Logging logging) throws IOException {
        //TODO send logs to another system.
        return "Dummy";
        // 1% of the time will throw exception
        if (RANDOM.nextInt(100) == 0) {
            throw new IOException("Dummy Exception");
        } else {
            return "Dummy";
        }
    }
```

chenqi0805 (approved):
Confirmed that by reverting the random error introduced in analytics-service, there will be no status code = 2 related to /logs. So nothing blocking AFAICT.

chenqi0805 (approved):

PR #168, opened 2020-11-20
Adding AbstractProcessor to record base metrics for processors

AustinTag:
*Issue #, if available:*

*Description of changes:*

- Adding AbstractProcessor class
- Using Micrometer to record metrics for records in, records out, and time elapsed in Abstract Processor
- Forced an HdrHistogram version in ES build.gradle since one of the ES dependencies pulls in an old version of this dependency

### DIFF
--- a/situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
+++ b/situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
@@ -1,0 +1,8 @@
+package com.amazon.situp.model.metrics;
+
+public class MetricNames {
+    public static final String RECORDS_IN = "recordsIn";
+    public static final String RECORDS_OUT = "recordsOut";
+    public static final String TIME_ELAPSED = "timeElapsed";
+    public static final String DELIMITER = ".";
+}
--- a/situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
+++ b/situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
@@ -1,8 +1,26 @@
 package com.amazon.situp.model.metrics;
 
 public class MetricNames {
+    /**
+     * Metric representing the ingress of records to a pipeline component.
+     * Applicable to processors and sinks
+     */
     public static final String RECORDS_IN = "recordsIn";
+
+    /**
+     * Metric representing the egress of records from a pipeline component.
+     * Applicable to processors and sources.
+     */
     public static final String RECORDS_OUT = "recordsOut";
+
+    /**
+     * Metric representing the time elapsed during execution of a pipeline component.
+     * Applicable to processors and sinks.
+     */
     public static final String TIME_ELAPSED = "timeElapsed";
+
+    /**
+     * Delimiter used to separate path components in metric names.
+     */
     public static final String DELIMITER = ".";
 }
--- a/situp-api/src/main/java/com/amazon/situp/model/processor/AbstractProcessor.java
+++ b/situp-api/src/main/java/com/amazon/situp/model/processor/AbstractProcessor.java
@@ -1,0 +1,48 @@
+package com.amazon.situp.model.processor;
+
+import com.amazon.situp.model.configuration.PluginSetting;
+import com.amazon.situp.model.record.Record;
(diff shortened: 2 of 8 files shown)


kowshikn:
The latest version makes it easy for the Plugin implementations but I am not sure if this is the best approach we have got. 

* Not sure i we will have issues in the `GET API` if we have prefix based metrics particularly if we use a backend like Promotheus. 
* Will this impact Pipeline-to-pipeline metrics? Any thoughts @yadavcbala ?
* How do we prevent users from not using the standard metrics? I agree this is something implementations should avoid but it would be good if we think of way we can avoid this?

I want to check if using a separate MetricRegistry for each pipeline and also provide each plugin its own registry will make this easier. Also use tags to identify each metric registry.  [Not talking about composite regsitry]

AustinTag:
So the default servlet for Prometheus supports getting metrics for a given list of metric names (but not by prefix). It will not be much extra work us to write an extension to support getting metrics by prefix. Having separate MeterRegistries for each pipeline and plugin will prevent us from using some of the Micrometer functionality, and require us to write a lot of new code to handle this. It is also against the conventions of Micrometer for us to keep many separate registries in the same application. At that point, we may as well not use the face of Micrometer.

situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
```diff
@@ -0,0 +1,8 @@
+package com.amazon.situp.model.metrics;
+
+public class MetricNames {
+    public static final String RECORDS_IN = "recordsIn";
+    public static final String RECORDS_OUT = "recordsOut";
+    public static final String TIME_ELAPSED = "timeElapsed";
+    public static final String DELIMITER = ".";
+}
```
kowshikn, line 8: nit: Add some documentation for each metrics. I believe these metrics are common for all plugins but some are applicable only to source and some to sink. So lets update them.

situp-api/src/main/java/com/amazon/situp/model/processor/AbstractProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+import com.amazon.situp.model.metrics.MetricNames;
+import com.amazon.situp.model.metrics.PluginMetrics;
+import com.amazon.situp.model.record.Record;
+import java.util.Collection;
+import java.util.StringJoiner;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Timer;
+
+public abstract class AbstractProcessor<InputRecord extends Record<?>, OutputRecord extends Record<?>> implements
```
kowshikn, line 13: s/Processor/Piper

situp-api/src/main/java/com/amazon/situp/model/processor/AbstractProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+import com.amazon.situp.model.configuration.PluginSetting;
+import com.amazon.situp.model.metrics.MetricNames;
+import com.amazon.situp.model.metrics.PluginMetrics;
+import com.amazon.situp.model.record.Record;
+import java.util.Collection;
+import java.util.StringJoiner;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Timer;
+
```
kowshikn, line 12: nit: Java Doc.

situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
```diff
@@ -0,0 +1,26 @@
+package com.amazon.situp.model.metrics;
+
+public class MetricNames {
```
kowshikn, line 3: Buffer metrics?

kowshikn (approved):

situp-api/src/main/java/com/amazon/situp/model/metrics/MetricNames.java
AustinTag: Sure, will add those when I add the buffer metrics

PR #182, opened 2020-11-30
Adding abstract implementations of Sink and Buffer

AustinTag:
*Issue #, if available:*

*Description of changes:*
Adding Abstract implementations of Buffer and Sink for testing.

Note: The checked exception in Buffer.write() required a little bit of extra code to handle in the abstract class. The other method for doing this would be to time the method, and use the Timer.record(Duration duration) to record the time. Both implementations required a little extra code, so I did not see a huge difference between the two.

### DIFF
--- a/situp-api/src/main/java/com/amazon/situp/model/buffer/AbstractBuffer.java
+++ b/situp-api/src/main/java/com/amazon/situp/model/buffer/AbstractBuffer.java
@@ -1,0 +1,87 @@
+package com.amazon.situp.model.buffer;
+
+import com.amazon.situp.model.configuration.PluginSetting;
+import com.amazon.situp.model.metrics.MetricNames;
+import com.amazon.situp.model.metrics.PluginMetrics;
+import com.amazon.situp.model.record.Record;
+import java.util.Collection;
+import java.util.concurrent.TimeoutException;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Timer;
+
+/**
+ * Abstract implementation of the Buffer interface to record boilerplate metrics
+ */
+public abstract class AbstractBuffer<T extends Record<?>> implements Buffer<T> {
+    protected final PluginMetrics pluginMetrics;
+    private final Counter recordsWrittenCounter;
+    private final Counter recordsReadCounter;
+    private final Counter writeTimeoutCounter;
+    private final Timer writeTimer;
+    private final Timer readTimer;
+
+    public AbstractBuffer(final PluginSetting pluginSetting) {
+        this.pluginMetrics = PluginMetrics.fromPluginSetting(pluginSetting);
+        this.recordsWrittenCounter = pluginMetrics.counter(MetricNames.RECORDS_WRITTEN);
+        this.recordsReadCounter = pluginMetrics.counter(MetricNames.RECORDS_READ);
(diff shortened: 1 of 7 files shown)


situp-api/src/main/java/com/amazon/situp/model/buffer/AbstractBuffer.java
```diff
@@ -0,0 +1,87 @@
...
+                try {
+                    doWrite(record, timeoutInMillis);
+                } catch (TimeoutException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+            recordsWrittenCounter.increment();
+        } catch (RuntimeException e) {
+            if(e.getCause() instanceof TimeoutException) {
+                writeTimeoutCounter.increment();
```
kowshikn, line 52: Can we not increment the counter in the catch statement above?

kowshikn (commented):
I know you are working in the branch, but heads up you are in for big merge issue as the  modules and package setting will no longer be referencing situp.

situp-api/src/main/java/com/amazon/situp/model/buffer/AbstractBuffer.java
```diff
@@ -0,0 +1,87 @@
+package com.amazon.situp.model.buffer;
```
yadavcbala, line 1: Everything currently is being renamed as per new legal name - This would have lot of conflicts

situp-api/src/main/java/com/amazon/situp/model/sink/AbstractSink.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /**
+     * Records metrics for ingress and time elapsed, while calling
+     * doOutput to perform the actual output logic
+     * @param records the records to write to the sink.
+     */
+    @Override
+    public void output(Collection<T> records) {
+        recordsInCounter.increment(records.size()*1.0);
```
yadavcbala, line 32: is this right place to update metrics - what happens if output fails but then we have already updated the metrics. Does `RECORDS_IN` is only measuring the records in for the sink and not the successful processing ?

yadavcbala, line 32: `records.size() * 1.0`

situp-api/src/main/java/com/amazon/situp/model/buffer/AbstractBuffer.java
AustinTag: Sure, that's fine.

situp-api/src/main/java/com/amazon/situp/model/sink/AbstractSink.java
AustinTag: This records records received by the processor

AustinTag: Increment takes a double

yadavcbala: was asking to add space :-)

yadavcbala (approved):
No blocking comments but as there is renaming happening in parallel, merging this in metrics branch and then to master will have merge conflicts.

situp-api/src/main/java/com/amazon/situp/model/sink/AbstractSink.java
AustinTag: oh, hah, somehow i misread that :P

PR #196, opened 2020-12-04
Resolves #194: Handle ES version conflict and resource already exists…

### LINKED ISSUE #194: Errors when running multiple instances of data-prepper in the sample app
Encountering ES client errors:

Example:
```
Caused by: java.lang.RuntimeException: method [PUT], host [https://node-0.example.com:9200], URI [/_opendistro/_ism/policies/raw-span-policy], status line [HTTP/1.1 409 Conflict]
{"error":{"root_cause":[{"type":"version_conflict_engine_exception","reason":"[raw-span-policy]: version conflict, document already exists (current version [1])","index_uuid":"IdnxdC94Qaa7CYJwSPESxg","shard":"0","index":".opendistro-ism-config"}],"type":"version_conflict_engine_exception","reason":"[raw-span-policy]: version conflict, document already exists (current version [1])","index_uuid":"IdnxdC94Qaa7CYJwSPESxg","shard":"0","index":".opendistro-ism-config"},"status":409}
        at com.amazon.dataprepper.plugins.sink.elasticsearch.ElasticsearchSink.<init>(ElasticsearchSink.java:62)
        ... 19 more
Caused by: org.elasticsearch.client.ResponseException: method [PUT], host [https://node-0.example.com:9200], URI [/_opendistro/_ism/policies/raw-span-policy], status line [HTTP/1.1 409 Conflict]
{"error":{"root_cause":[{"type":"version_conflict_engine_exception","reason":"[raw-span-policy]: version conflict, document already exists (current version [1])","index_uuid":"IdnxdC94Qaa7CYJwSPESxg","shard":"0","index":".opendistro-ism-config"}],"type":"version_conflict_engine_exception","reason":"[raw-span-policy]: version conflict, document already exists (current version [1])","index_uuid":"IdnxdC94Qaa7CYJwSPESxg","shard":"0","index":".opendistro-ism-config"},"status":409}
....
2523 [main] ERROR com.amazon.dataprepper.parser.PipelineParser  – Construction of pipeline components failed, skipping building of pipeline [raw-pipeline]

```

So far I've hit issues in ElasticsearchSink.java and IndexStateManagement.java, both around the lowLevelClient performing requests. Need to handle these errors gracefully instead of failing the pipeline initialization completely.

wrijeff:
… exceptions on sink startup.

*Issue #, if available:* #194 

*Description of changes:*
There are race conditions when multiple ES sinks start up for the first time, resulting in multiple attempts to put the same policy or index to the same ES cluster. 

I couldn't find a graceful way to parse the Elasticsearch exception types other than by reading the exception message contents :-/ (Here's a [stackoverflow question](https://stackoverflow.com/questions/52345980/elasticsearch-catching-exceptions-with-the-rest-high-level-client) where someone else ran into the same problem). I found [a related Elasticsearch issue](https://github.com/elastic/elasticsearch/issues/37020) but wasn't able to identify the fix.

### DIFF
--- a/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
+++ b/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
@@ -5,6 +5,7 @@ import com.amazon.dataprepper.model.annotations.DataPrepperPlugin;
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.record.Record;
 import com.amazon.dataprepper.model.sink.Sink;
+import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.action.DocWriteRequest;
 import org.elasticsearch.action.admin.indices.alias.Alias;
 import org.elasticsearch.action.admin.indices.alias.get.GetAliasesRequest;
@@ -15,7 +16,6 @@ import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.indices.CreateIndexRequest;
 import org.elasticsearch.client.indices.GetIndexRequest;
 import org.elasticsearch.client.indices.PutIndexTemplateRequest;
-
 import org.elasticsearch.common.unit.ByteSizeUnit;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
 import org.elasticsearch.common.xcontent.NamedXContentRegistry;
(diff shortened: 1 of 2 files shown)


kowshikn:
I thought the race condition happens for index state policy creation i.e createIndexTemplate. Is it not?

wrijeff:
> I thought the race condition happens for index state policy creation i.e createIndexTemplate. Is it not?

There were 2 places where I ran into errors when starting up the sample app w/ 2 Prepper instances - checkAndCreatePolicy() and checkAndCreateIndex(). Though any of the "check and create" approaches are subject to this race condition because the check/create actions don't occur in isolation.

I don't recall any problems with createIndexTemplate() while debugging this issue, maybe putTemplate() doesn't fail when called multiple times?

chenqi0805:
> maybe putTemplate() doesn't fail when called multiple times?

I think ES just "overwrites" the index template by incrementing the version number of the template.

kowshikn:
@chenqi0805 @wrijeff Should we also handle the exception in check and create template?

chenqi0805:
> @chenqi0805 @wrijeff Should we also handle the exception in check and create template?

So just to clarify first

1. There is no check and create but only createIndexTemplate called.
2. createIndexTemplate will not throw exception even if the template already exists in ES. It only increment the version number of the template if called multiple times. 

For now createIndexTemplate only throws IOException, which is handled when calling start()

```
public ElasticsearchSink(final PluginSetting pluginSetting) {
    this.esSinkConfig = ElasticsearchSinkConfiguration.readESConfig(pluginSetting);
    this.bulkSize = ByteSizeUnit.MB.toBytes(esSinkConfig.getIndexConfiguration().getBulkSize());
    this.indexType = esSinkConfig.getIndexConfiguration().getIndexType();
    this.documentIdField = esSinkConfig.getIndexConfiguration().getDocumentIdField();
    try {
      start();
    } catch (final IOException e) {
      throw new RuntimeException(e.getMessage(), e);
    }
  }
```
Do you mean we should handle it differently?

kowshikn (approved):

chenqi0805 (approved):
LGTM.

Issue #200, opened 2020-12-07
Peer Forwarder Plugin


kowshikn:
The trace analytics feature is the first pipeline that Data Prepper will offer. This feature is requires spans belonging to the same trace workflow to be routed to the same host so we can process and make decision on the entire trace workflow.

The end state is we have a processor plugin,

```
processor:
    - peer_forwarder:
       discovery.mode: [] 
       peer.host: []
       time_out: 300
       span_agg_count: 48 
```

The plugin should have an option to auto discovery/list of peers and support the routing.

Items:

- [x] consistent hashing algorithm
- [x] Discovery approach
- [x] Integrate with metrics(Austin already has the abstract classes)
   - [x] Inherit common metrics abstract classes
   - [x] Add custom metrics by introducing timer, counter, etc in constructor
- [x] Integration tests that ensure we are routing them correctly and detecting service map relationship.
   - [x] Update e2e test to include at least 2 data prepper instances to receive traces.
   - [x] Integration test for discovery mode #224 
- [x] SSL/TLS support
- [x] Documentation on how to configure various options

chenqi0805:
Requirements for hashing:

1. support add/remove dataPrepper IP address (Or support rebuild the ring when there is change in IPs)
2. Threadsafety? Depends on how discovery works, but IMO it is likely to be called by an async process.
3. Reroute as few requests as possible whenever there is an update on dataPrepper IP addresses?

kowshikn:
* Do we have plans to add ec2 based discovery?
* Do we have plans to test and documentation for users to make the DNS based setup work in non Docker/K8s setup?

wrijeff:
> * Do we have plans to add ec2 based discovery?
> * Do we have plans to test and documentation for users to make the DNS based setup work in non Docker/K8s setup?

Good idea, will create issues for each. Can prioritize higher/lower depending on what other tasks are in the backlog.

wrijeff:
Closing this parent issue. Integ tests to be handled in #224

PR #206, opened 2020-12-08
Adding DNS container to trace-analytics-sample-app.

wrijeff:
*Issue #, if available:* #200 

*Description of changes:*
Adding a small DNS service to the docker-compose example, to be used in the multi-host DNS discovery example.

Tested with `docker compose up -d` and confirming data still propagates all the way to ES:
>15846 [raw-pipeline-process-worker-3-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker  –  raw-pipeline Worker: Processing 7 records from buffer

### DIFF
--- a/examples/dns/dnsmasq.conf
+++ b/examples/dns/dnsmasq.conf
@@ -1,0 +1,3 @@
+log-queries
+server=127.0.0.11
+addn-hosts=/etc/hosts.prepper
--- a/examples/dns/dnsmasq.conf
+++ b/examples/dns/dnsmasq.conf
@@ -1,3 +1,2 @@
-log-queries
-server=127.0.0.11
+server=127.0.0.11 # Embedded Docker DNS
 addn-hosts=/etc/hosts.prepper
--- a/examples/trace-analytics-sample-app/opentelemetry-collector/otel-collector-dns-config.yml
+++ b/examples/trace-analytics-sample-app/opentelemetry-collector/otel-collector-dns-config.yml
@@ -1,0 +1,17 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:55680
+
+exporters:
+  otlp/2:
+    endpoint: prepper-cluster:21890
+    insecure: true
+  logging:
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [logging, otlp/2]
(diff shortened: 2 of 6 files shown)


examples/trace-analytics-sample-app/opentelemetry-collector/otel-collector-dns-config.yml
```diff
@@ -0,0 +1,17 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:55680
+
+exporters:
+  otlp/2:
+    endpoint: prepper-cluster:21890
```
chenqi0805, line 9: QUES: just curious. Will 10.10.2.1:21890 also work?

examples/dns/dnsmasq.conf
```diff
@@ -0,0 +1,3 @@
+log-queries
+server=127.0.0.11
```
chenqi0805, line 2: QUES: Is the IP typo? I actually do not quite understand what this flag means. In the doc it says `Specify IP address of upstream servers directly`

chenqi0805 (commented):
Just some of my own questions on this setup. Could you also add some description about how DNS container works in the README?

wrijeff:
> Just some of my own questions on this setup. Could you also add some description about how DNS container works in the README?

Sure, will add documentation.

examples/trace-analytics-sample-app/opentelemetry-collector/otel-collector-dns-config.yml
wrijeff: Correct, any of the following will work.
* `10.10.2.1` - the address of the prepper instance
* `data-prepper` - provided by [docker-compose](https://docs.docker.com/compose/networking/) (hostname = container name)
* `prepper-cluster` - defined in dnsmasq configuration

In this example I'm pointing the otel exporter at `prepper-cluster` so we can easily test with multiple hosts by adding entries to the dnsmasq hosts.prepper file.

examples/dns/dnsmasq.conf
wrijeff: Not a typo (I can add a comment). 

Docker's internal embedded DNS happens to be at `127.0.0.11`, so by setting this as an upstream server we are telling dnsmasq to relay requests to the Docker's DNS for names which dnsmasq can't resolve.

chenqi0805 (approved):

wrijeff:
Closing as project structure is changing - will most likely include these changes in a follow-up PR.

PR #211, opened 2020-12-09
MAINT: make PeerForwarder extends metric compatible abstract class

chenqi0805:
*Issue #, if available:*
Contributes to #200 

*Description of changes:*
Manifested inline.


wrijeff (approved):
Seems fine (though I'm not 100% up to speed on the metrics changes)

kowshikn:
The goal of this PR is just to rename and get the benefit of default metrics. There should be a separate PR to add specific/custom metrics for Peer forwarder plugin. Is that correct @chenqi0805 ?

LGTM.

kowshikn (approved):

chenqi0805:
> There should be a separate PR to add specific/custom metrics for Peer forwarder plugin. Is that correct @chenqi0805 ?

Yes. Should be a separate PR. Need to sync up on what those metrics are.

Issue #212, opened 2020-12-09
Use the latest image in our sample to reduce startup time.


kowshikn:
In our examples, we will use the docker image instead of building the projects.

kowshikn:
Note: `latest` doesn't work, so we are using the specific version.

wrijeff:
Merged #213

PR #213, opened 2020-12-09
"Use the docker image in sample app"

kowshikn:
Issue #212 

Using `restart` for data prepper to start repeatedly till ODFE becomes available.


examples/adot/docker-compose.yml
```diff
@@ -27,16 +27,12 @@ services:
...
       - my_network
   data-prepper:
+    restart: unless-stopped
     container_name: data-prepper
-    build:
-      context: ../..
-      dockerfile: examples/Dockerfile
-    working_dir: /app
-    command: sh data-prepper-wait-for-odfe-and-start.sh
+    image: opendistroforelasticsearch/data-prepper:v0.7.0-alpha
```
yadavcbala, line 32: this is staging repository - you can as well use production after Dec14

yadavcbala (approved):

yadavcbala:
Also, since you are using the image for data-prepper, the examples will not be able to use any non-released features - I am assuming you want it to be like this

examples/adot/docker-compose.yml
kowshikn: Okay. Will follow up with another PR on 14th.

chenqi0805 (approved):
LGTM. Same ques as @yadavcbala . I am OK with the change. But do we want customer to run examples on releases or on the bleeding edge of the data-prepper.

kowshikn:
> LGTM. Same ques as @yadavcbala . I am OK with the change. But do we want customer to run examples on releases or on the bleeding edge of the data-prepper.

Releases, because we want them to pull the latest approved version and also avoid gradle build etc.

PR #218, opened 2020-12-11
Updating OtelTraceSourceTests

AustinTag:
*Issue #, if available:*

https://github.com/opendistro-for-elasticsearch/Data-Prepper/issues/214

*Description of changes:*

- Updating armeria WebClient.execute() call to call join(), so that the future is run
- Updating test to assert a 415 (media type unsupported) error, which is what is expected
- Removing extra test for http, since this is no longer needed


kowshikn:
nit: Can keep and check for 415 on HTTPBytes tests as well? The reason when we allow http, we will support both JSON and Bytes.

kowshikn (approved):

kowshikn:
Great catch @AustinTag :)

chenqi0805 (approved):
LGTM except for the failed test case in CI

PR #226, opened 2020-12-14
Starting data prepper server when initializing the data prepper instance

AustinTag:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/Data-Prepper/issues/201

*Description of changes:*

- Starting a DataPrepperServer when instantiating DataPrepper instance
- Exposing metrics port on docker container for integ test
- Manually tested getting metrics with integration test
   - Actually verifying metrics in the integ test will be cumbersome, TODO later
- Taking in the server port as configuration will be handled in a later task, as this will require a little bit of a refactor in how we parse configuration

### DIFF
--- a/data-prepper-core/integrationTest.gradle
+++ b/data-prepper-core/integrationTest.gradle
@@ -55,6 +55,7 @@ task createDataPrepperDockerFile(type: Dockerfile) {
     destFile = project.file('build/docker/Dockerfile')
     from("adoptopenjdk/openjdk14:jre-14.0.1_7-alpine")
     exposePort(21890)
+    exposePort(4900)
     workingDir("/app")
     copyFile("build/libs/${jar.archiveName}", "/app/data-prepper.jar")
     copyFile("src/integrationTest/resources/pipeline.yml", "/app/pipeline.yml")
@@ -71,7 +72,7 @@ task buildDataPrepperDockerImage(type: DockerBuildImage) {
 task createDataPrepperDockerContainer(type: DockerCreateContainer) {
     dependsOn buildDataPrepperDockerImage
     dependsOn createDataPrepperNetwork
-    hostConfig.portBindings = ['21890:21890']
+    hostConfig.portBindings = ['21890:21890', '4900:4900']
     hostConfig.network = createDataPrepperNetwork.getNetworkName()
     targetImageId buildDataPrepperDockerImage.getImageId()
 }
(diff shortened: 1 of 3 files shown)


data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
```diff
@@ -15,11 +19,14 @@
  */
 public class DataPrepper {
     private static final Logger LOG = LoggerFactory.getLogger(DataPrepper.class);
+    private static final int SERVER_PORT = 4900;
```
kowshikn, line 22: Maybe we should make this port also configurable?

data-prepper-core/integrationTest.gradle
```diff
@@ -71,7 +72,7 @@ task buildDataPrepperDockerImage(type: DockerBuildImage) {
 task createDataPrepperDockerContainer(type: DockerCreateContainer) {
     dependsOn buildDataPrepperDockerImage
     dependsOn createDataPrepperNetwork
-    hostConfig.portBindings = ['21890:21890']
+    hostConfig.portBindings = ['21890:21890', '4900:4900']
```
kowshikn, line 75: Is the test using the new port?

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
kowshikn: (Just checked the description)
I agree we need to have a config specific to DataPrepper. This is probably something that should be part of the config file for Data Prepper. We should not use the pipeline.yml file.

kowshikn (approved):

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
```diff
@@ -60,6 +78,7 @@ public void shutdown() {
         transformationPipelines.forEach((name, pipeline) -> {
             pipeline.shutdown();
         });
+        dataPrepperServer.stop();
```
yadavcbala, line 81: nit: what happens when customer wants to shutdown 1 of the 1 pipeline running using its name - I think the disconnect is in the existing code. [just have this in back of ur mind]

yadavcbala (approved):

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
AustinTag: Yeah, I think we need to define what the behavior should be. Potentially the server can still run if all the pipelines are shut down.

PR #229, opened 2020-12-14
#200: Updated PeerListProvider interface to extend Listenable, update…

wrijeff:
…d implementations.

*Issue #, if available:* #200 

*Description of changes:*
The hash ring needs to be rebuilt whenever the list of peers changes. Hooked in to Armeria's Listenable interface for this purpose.
* Updated PeerListProvider interface to extend Armeria's Listenable interface
* Updated static and DNS implementations

Testing:
Started a cluster w/ DNS discovery, scaled from 1 node to 3 and observed logs:
```
...
1113 [main] INFO  com.amazon.dataprepper.plugins.processor.peerforwarder.HashRing  – Building hash ring with endpoints: [172.17.0.6]
...
5783 [armeria-boss-http-*:21890] INFO  com.linecorp.armeria.server.Server  – Serving HTTP at /0.0.0.0:21890 - http://127.0.0.1:21890/
5783 [main] INFO  com.amazon.dataprepper.plugins.source.oteltrace.OTelTraceSource  – Started otel_trace_source...
5783 [main] INFO  com.amazon.dataprepper.pipeline.Pipeline  – Pipeline [entry-pipeline] - Submitting request to initiate the pipeline processing
5884 [entry-pipeline-process-worker-1-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker  –  entry-pipeline Worker: No records received from buffer
7914 [raw-pipeline-process-worker-3-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker  –  raw-pipeline Worker: No records received from buffer
15534 [entry-pipeline-process-worker-1-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker  –  entry-pipeline Worker: Processing 1 records from buffer
15640 [service-map-pipeline-process-worker-2-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker  –  service-map-pipeline Worker: Processing 1 records from buffer
...
61119 [armeria-common-worker-epoll-2-1] INFO  com.amazon.dataprepper.plugins.processor.peerforwarder.HashRing  – Building hash ring with endpoints: [172.17.0.11, 172.17.0.12, 172.17.0.6]

```

### DIFF
--- a/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
+++ b/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
@@ -17,12 +17,18 @@ import java.util.Optional;
 import java.util.TreeMap;
 import java.util.function.Consumer;
 
+/**
+ * Consistent hashing implementation used to map trace IDs to Data Prepper hosts.
+ * See https://en.wikipedia.org/wiki/Consistent_hashing for more information.
+ */
 @NotThreadSafe
 public class HashRing implements Consumer<List<Endpoint>> {
     private static final Logger LOG = LoggerFactory.getLogger(HashRing.class);
     private static final String MD5 = "MD5";
 
+    /* Number of virtual nodes per Data Prepper host to be present on the hash ring */
     private final int numVirtualNodes;
+
     private final PeerListProvider peerListProvider;
 
     private TreeMap<BigInteger, String> hashServerMap = new TreeMap<>();
--- a/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
+++ b/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
@@ -1,19 +1,25 @@
 package com.amazon.dataprepper.plugins.processor.peerforwarder;
 
 import com.amazon.dataprepper.plugins.processor.peerforwarder.discovery.PeerListProvider;
+import com.linecorp.armeria.client.Endpoint;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.concurrent.NotThreadSafe;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.function.Consumer;
 
 @NotThreadSafe
-public class HashRing {
+public class HashRing implements Consumer<List<Endpoint>> {
+    private static final Logger LOG = LoggerFactory.getLogger(HashRing.class);
     private static final String MD5 = "MD5";
 
     private final int numVirtualNodes;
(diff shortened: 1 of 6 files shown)


data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
```diff
@@ -27,6 +33,8 @@ public HashRing(final PeerListProvider peerListProvider, final int numVirtualNod
         this.numVirtualNodes = numVirtualNodes;
 
         buildHashServerMap();
+
+        peerListProvider.addListener(this);
```
chenqi0805, line 37: QUES: I do not have a good answer for solutions. But would like to raise concern here

1. Will there be concurrency issue on reading and writing the state of the instance ?
2. Publish `this` in the constructor might add the risk of publishing the instance in an invalid state. Might not be the issue now as this is the last line but for future maintenance.

chenqi0805 (commented):
Raised some questions. Otherwise LGTM.

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
wrijeff: 1. I don't expect concurrency to be an issue as we don't expect the list to change at a high frequency. `buildHashServerMap()` creates a new map variable then assigns it to the class variable once it's built completely, so no caller will be reading from a "partially built" map.
2. Fair point - we're making the assumption that there's a very low chance of the object being called by the Listenable before the object has finished being created. I'll leave it for now, but could split this out of the constructor into its own method in a following PR.

chenqi0805 (approved):

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
```diff
@@ -1,19 +1,25 @@
...
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.function.Consumer;
 
 @NotThreadSafe
-public class HashRing {
+public class HashRing implements Consumer<List<Endpoint>> {
```
AustinTag, line 21: Nit: probably want javadoc in this class

AustinTag (commented):
Something that stands out to me is that there is some considerable logic added, but no tests are added to confirm that the new code will function as expected. Is the logic/functionality covered by existing tests?

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/HashRing.java
wrijeff: Will add 👍

wrijeff:
> Something that stands out to me is that there is some considerable logic added, but no tests are added to confirm that the new code will function as expected. Is the logic/functionality covered by existing tests?

Fair point - we have 2 new methods which intentionally do nothing, and 2 new methods which make 1-line calls to an Armeria object (which should be covered by its own unit tests). I can add tests to at least verify the Armeria object was called correctly however.

AustinTag (approved):

PR #236, opened 2020-12-15
Adding "/list" and "/shutdown" APIs to DataPrepperServer

AustinTag:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/Data-Prepper/issues/233

*Description of changes:*
- Adding HttpHandlers for List and Shutdown APIs
- Accepting pointer to DataPrepper into the DataPrepperServer constructor
- Adding tests for the 2 new handlers

Note: The PipelineListing object will contain more information about the pipeline (such as isRunning) in the future, when that information is available from the pipeline.

### DIFF
--- a/data-prepper-core/build.gradle
+++ b/data-prepper-core/build.gradle
@@ -25,7 +25,7 @@ dependencies {
     implementation "io.micrometer:micrometer-core:1.5.1"
     implementation "io.micrometer:micrometer-registry-prometheus:1.5.1"
     testImplementation "org.hamcrest:hamcrest:${versionMap.hamcrest}"
-    testImplementation "org.mockito:mockito-all:1.10.19"
+    testImplementation "org.mockito:mockito-core:${versionMap.mockito}"
 }
 
 jar {
--- a/data-prepper-core/build.gradle
+++ b/data-prepper-core/build.gradle
@@ -25,6 +25,7 @@ dependencies {
     implementation "io.micrometer:micrometer-core:1.5.1"
     implementation "io.micrometer:micrometer-registry-prometheus:1.5.1"
     testImplementation "org.hamcrest:hamcrest:${versionMap.hamcrest}"
+    testImplementation "org.mockito:mockito-all:1.10.19"
 }
 
 jar {
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
@@ -1,5 +1,6 @@
 package com.amazon.dataprepper.pipeline.server;
 
+import com.amazon.dataprepper.DataPrepper;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import com.sun.net.httpserver.HttpServer;
@@ -11,14 +12,18 @@ import com.sun.net.httpserver.HttpServer;
 public class DataPrepperServer {
 
     private final HttpServer server;
+    private final DataPrepper dataPrepper;
 
-    public DataPrepperServer(final int port) {
+    public DataPrepperServer(final int port, final DataPrepper dataPrepper) {
+        this.dataPrepper = dataPrepper;
         try {
             server = HttpServer.create(
                     new InetSocketAddress(port),
                     0
             );
             server.createContext("/metrics/prometheus", new PrometheusMetricsHandler());
+            server.createContext("/list", new ListPipelinesHandler(dataPrepper));
+            server.createContext("/shutdown", new ShutdownHandler(dataPrepper));
         } catch (IOException e) {
             throw new RuntimeException("Failed to create server", e);
         }
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
@@ -12,10 +12,8 @@ import com.sun.net.httpserver.HttpServer;
 public class DataPrepperServer {
 
     private final HttpServer server;
-    private final DataPrepper dataPrepper;
 
     public DataPrepperServer(final int port, final DataPrepper dataPrepper) {
-        this.dataPrepper = dataPrepper;
         try {
             server = HttpServer.create(
                     new InetSocketAddress(port),
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
@@ -1,0 +1,54 @@
+package com.amazon.dataprepper.pipeline.server;
+
+import com.amazon.dataprepper.DataPrepper;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.sun.net.httpserver.HttpExchange;
+import com.sun.net.httpserver.HttpHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * HttpHandler to handle requests for listing pipelines running on the data prepper instance
+ */
+public class ListPipelinesHandler implements HttpHandler {
+
+    private final DataPrepper dataPrepper;
+    private final static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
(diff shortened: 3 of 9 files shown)


data-prepper-core/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
     implementation "io.micrometer:micrometer-core:1.5.1"
     implementation "io.micrometer:micrometer-registry-prometheus:1.5.1"
     testImplementation "org.hamcrest:hamcrest:${versionMap.hamcrest}"
+    testImplementation "org.mockito:mockito-all:1.10.19"
```
wrijeff, line 28: Might be time to wrangle all the mockito versions and stick them in the version map. 

[2 occurrences so far](https://github.com/opendistro-for-elasticsearch/Data-Prepper/search?q=org.mockito+filename%3A*.gradle), I'm guilty of adding one of them 😅

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
```diff
@@ -0,0 +1,54 @@
...
+                    .collect(Collectors.toList());
+            final byte[] response = OBJECT_MAPPER.writeValueAsString(Collections.singletonMap("pipelines", pipelines)).getBytes();
+            exchange.getResponseHeaders().add("Content-Type", "text/plain; charset=UTF-8");
+            exchange.sendResponseHeaders(200, response.length);
+            exchange.getResponseBody().write(response);
+            exchange.getResponseBody().close();
+        } catch (Exception e) {
+            LOG.error("Caught exception listing pipelines", e);
+            exchange.sendResponseHeaders(500, 0);
+            exchange.getResponseBody().close();
```
wrijeff, line 51: nit: Not that it's any functionally different, but these `close()` lines could be moved out into a finally block.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -12,13 +13,15 @@
 
     private final HttpServer server;
 
-    public DataPrepperServer(final int port) {
+    public DataPrepperServer(final int port, final DataPrepper dataPrepper) {
```
wrijeff, line 16: Do we want to enforce non-null here (`Objects.requireNonNull`)? I'm asking because I saw this line below in a test file which threw me off a bit:
```
final DataPrepperServer dataPrepperServer = new DataPrepperServer(8080, null);
```

Could also be a question for the larger project itself - would these checks be useful in most constructors?

data-prepper-core/src/test/java/com/amazon/dataprepper/server/DataPrepperServerTest.java
```diff
@@ -50,6 +58,76 @@ public void testScrapeFailure() throws IOException, InterruptedException, URISyn
         dataPrepperServer.stop();
     }
 
+    @Test
+    public void testListPipelines() throws URISyntaxException, IOException, InterruptedException {
+        setRegistry(new PrometheusRegistryMockScrape(PrometheusConfig.DEFAULT, ""));
+        final DataPrepper mockDataPrepper = Mockito.mock(DataPrepper.class);
```
wrijeff, line 64: nit: Shared setup code like this which _always_ is executed could be brought up into a `@Before` method to make the tests themselves more concise.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    public void handle(HttpExchange exchange) throws IOException {
+        try {
+            List<PipelineListing> pipelines = dataPrepper.getTransformationPipelines()
+                    .entrySet()
+                    .stream()
+                    .map(entry -> new PipelineListing(entry.getKey()))
+                    .collect(Collectors.toList());
+            final byte[] response = OBJECT_MAPPER.writeValueAsString(Collections.singletonMap("pipelines", pipelines)).getBytes();
+            exchange.getResponseHeaders().add("Content-Type", "text/plain; charset=UTF-8");
+            exchange.sendResponseHeaders(200, response.length);
```
wrijeff, line 45: nit: I'm kind of surprised our dependencies don't have a response code enum defined 🤔. Could throw together a quick constants file if we think it could bring value (e.g. keeping values consistent across classes and test classes).

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ShutdownHandler.java
```diff
@@ -0,0 +1,34 @@
...
+    private final static Logger LOG = LoggerFactory.getLogger(ShutdownHandler.class);
+
+    public ShutdownHandler(final DataPrepper dataPrepper) {
+        this.dataPrepper = dataPrepper;
+    }
+
+    @Override
+    public void handle(HttpExchange exchange) throws IOException {
+        try {
+            dataPrepper.shutdown();
```
wrijeff, line 25: Just curious - if dataPrepper is already shutdown, will we return a 200 or 500?

data-prepper-core/src/test/java/com/amazon/dataprepper/server/DataPrepperServerTest.java
```diff
@@ -50,6 +58,76 @@ public void testScrapeFailure() throws IOException, InterruptedException, URISyn
...
+                .GET().build();
+        HttpResponse response = HttpClient.newHttpClient().send(request, HttpResponse.BodyHandlers.ofString());
+        final String expectedResponse = OBJECT_MAPPER.writeValueAsString(
+                Collections.singletonMap("pipelines", Arrays.asList(
+                        Collections.singletonMap("name", pipelineName)
+                ))
+        );
+        Assert.assertEquals(200, response.statusCode());
+        Assert.assertEquals(expectedResponse, response.body());
+        dataPrepperServer.stop();
```
wrijeff, line 82: nit: Similar comment about common code - could this be brought up into an `@After` block to ensure shutdown? Or at the very least put in a finally block?

wrijeff (commented):
All comments or questions are minor/nitpicks 👍

data-prepper-core/build.gradle
AustinTag: Yep, I can do that!

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
AustinTag: Ah, good call. Will do

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
AustinTag: Hmm, yeah. It's probably something that should be done at a larger level.

In this case, the constructor is just passing that reference to the constructors of some of the handlers, so maybe the NonNull requirement would make more sense in those constructors (because its not used explicitly here).

chenqi0805 (approved):
LGTM

data-prepper-core/src/test/java/com/amazon/dataprepper/server/DataPrepperServerTest.java
AustinTag: Yeah, it will need to be overridden in the prometheus specific tests, but I will add this in to an @Before function for the rest of the tests.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ListPipelinesHandler.java
AustinTag: Yeah, I'll poke around for a response code enum from sun.net or the java package. There is one from armeria, but its not worth bringing that whole thing in just for that.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/ShutdownHandler.java
AustinTag: So once dataPrepper is fully shut down, then the server will shut down as well.

data-prepper-core/src/test/java/com/amazon/dataprepper/server/DataPrepperServerTest.java
AustinTag: Hmm yeah, so the stop() function can be moved into a finally.

wrijeff (approved):
👍

Issue #239, opened 2020-12-16
Enforce code coverage standards across project


AustinTag:
A code coverage of 65% is far too low for our project at this stage. In addition, it seems that the jacoco code coverage is not being run on the many subprojects in the plugins project. Reports only seem to be generated for data-prepper-api,  data-prepper-core, and data-prepper-plugins/common. 

- [x] data-prepper-api
- [x] data-prepper-core
- [x] data-prepper-plugins/common
- [x] data-prepper-plugins/elasticsearch
- [ ] data-prepper-plugins/lmdb-processor-state
- [x] data-prepper-plugins/map-db-processor-state
- [x] data-prepper-plugins/otel-trace-raw-processor
- [x] data-prepper-plugins/otel-trace-source
- [x] data-prepper-plugins/peer-forwarder
- [x] data-prepper-plugins/service-map-stateful

AustinTag:
We need to have at least 95% coverage across the project, and should strive for 100%. There are very few cases where we should not be able to hit a line of code in testing.

yadavcbala:
Agreed - We already have the enforcement enabled for core projects but since they did not have 95% - It was set to lower levels

1. [Project level](https://github.com/opendistro-for-elasticsearch/Data-Prepper/blob/master/build.gradle#L25-L51) - _This enforces only core projects which are core, api and plugins:common_
2. [Core](https://github.com/opendistro-for-elasticsearch/Data-Prepper/blob/master/data-prepper-core/build.gradle#L50)
3. [Api this is at 100%](https://github.com/opendistro-for-elasticsearch/Data-Prepper/blob/master/data-prepper-api/build.gradle#L13)
4. [Plugins:Common](https://github.com/opendistro-for-elasticsearch/Data-Prepper/blob/master/data-prepper-plugins/common/build.gradle#L18)

But yeah we can easily get better than 65% in plugins and better than 60% in core

AustinTag:
Yeah, all plugins have unit testing at this point. But most are probably sitting around 70% or something. I think we should get the coverage requirement up asap so that (1) we know that we are testing well, and (2) new code can't be added without being thoroughly tested.

kowshikn:
@AustinTag @chenqi0805 - Whats the update on this? This should be closed before our March release.

AustinTag:
I think all of the plugins other than the lmdb state plugin are at at least 90% now

kowshikn:
[...]

PR #246, opened 2020-12-17
Applied the jacoco code coverage gradle plugin to all subprojects.

wrijeff:
*Issue #, if available:* #239

*Description of changes:*
Bringing up the jacoco plugin to the top level so that all projects will have access to coverage reports. Will eventually enable enforcement, but we need to establish baselines first.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ allprojects {
 
 subprojects {
     apply plugin: 'java'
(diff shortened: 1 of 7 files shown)


yadavcbala (approved):

chenqi0805 (approved):
LGTM!

PR #247, opened 2020-12-17
ADD: custom metrics in peer-forwarder

chenqi0805:
*Issue #, if available:*
#200 

*Description of changes:*
- ~Added forwardRequestErrorCounter as a custom metric in PeerForwarder~
- ~Added a test case to cover the metric~

This PR 

- Add an overload of gauge API in PluginMetrics
- Introduced peerforwarder custom metrics as in the checklist
  - [x] Timer
     - [x] Timer on forwarding each request per target peer IP
  - [x] Counter
     - [x] Counter on number of forwarded requests per target peer IP
     - [x] Counter on number of forward request errors per target peer IP
  - [x] Gauge
     - [x] Number of active peer IPs returned by PeerProviderList
- Refined test cases on PeerForwarder and verified metrics values

### DIFF
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
@@ -9,6 +9,10 @@ import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Timer;
 
+/**
+ * Provides reference to APIs that register timer, counter, gauge into global registry.
+ * See https://micrometer.io/docs/concepts#_registry
+ */
 public class PluginMetrics {
 
     private final String metricsPrefix;
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
@@ -2,6 +2,8 @@ package com.amazon.dataprepper.metrics;
 
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import java.util.StringJoiner;
+import java.util.function.ToDoubleFunction;
+
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Metrics;
@@ -41,7 +43,11 @@ public class PluginMetrics {
     }
 
     public <T extends Number> T gauge(final String name, T number) {
-        return Metrics.gauge(name, number);
+        return Metrics.gauge(getMeterName(name), number);
+    }
+
+    public <T> T gauge(String name, T obj, ToDoubleFunction<T> valueFunction) {
+        return Metrics.gauge(getMeterName(name), obj, valueFunction);
     }
 
     private String getMeterName(final String name) {
--- a/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
+++ b/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
@@ -133,17 +133,17 @@ public class PeerForwarder extends AbstractPrepper<Record<ExportTraceServiceRequ
                                 final List<Record<ExportTraceServiceRequest>> localBuffer) {
         if (client != null) {
             final String peerIp = client.getChannel().authority();
+            final Timer forwardRequestTimer = forwardRequestTimers.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.timer(String.format("%s:%s", FORWARD_REQUEST_LATENCY_PREFIX, ip)));
+            final Counter forwardedRequestCounter = forwardedRequestCounters.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_SUCCESS_PREFIX, ip)));
+            final Counter forwardRequestErrorCounter = forwardRequestErrorCounters.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_ERRORS_PREFIX, ip)));
             try {
-                final Timer forwardRequestTimer = forwardRequestTimers.computeIfAbsent(
-                        peerIp, ip -> pluginMetrics.timer(String.format("%s:%s", FORWARD_REQUEST_LATENCY_PREFIX, ip)));
                 forwardRequestTimer.record(() -> client.export(request));
-                final Counter forwardedRequestCounter = forwardedRequestCounters.computeIfAbsent(
-                        peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_SUCCESS_PREFIX, ip)));
                 forwardedRequestCounter.increment();
             } catch (Exception e) {
                 LOG.error(String.format("Failed to forward the request:\n%s\n", request.toString()));
-                final Counter forwardRequestErrorCounter = forwardRequestErrorCounters.computeIfAbsent(
-                        peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_ERRORS_PREFIX, ip)));
                 forwardRequestErrorCounter.increment();
                 localBuffer.add(new Record<>(request));
             }
--- a/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
+++ b/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
@@ -1,11 +1,13 @@
 package com.amazon.dataprepper.plugins.processor.peerforwarder;
 
+import com.amazon.dataprepper.metrics.PluginMetrics;
 import com.amazon.dataprepper.model.PluginType;
 import com.amazon.dataprepper.model.annotations.DataPrepperPlugin;
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.processor.AbstractPrepper;
 import com.amazon.dataprepper.model.record.Record;
 import com.amazon.dataprepper.plugins.processor.peerforwarder.discovery.StaticPeerListProvider;
+import io.micrometer.core.instrument.Counter;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
 import io.opentelemetry.proto.collector.trace.v1.TraceServiceGrpc;
 import io.opentelemetry.proto.trace.v1.ResourceSpans;
@@ -24,11 +26,14 @@ import java.util.Map;
 
 @DataPrepperPlugin(name = "peer_forwarder", type = PluginType.PROCESSOR)
 public class PeerForwarder extends AbstractPrepper<Record<ExportTraceServiceRequest>, Record<ExportTraceServiceRequest>> {
+    public static final String FORWARD_REQUEST_ERRORS = "forwardRequestErrors";
+
     private static final Logger LOG = LoggerFactory.getLogger(PeerForwarder.class);
 
     private final HashRing hashRing;
     private final PeerClientPool peerClientPool;
     private final int maxNumSpansPerRequest;
+    private final Counter forwardRequestErrorCounter;
 
     public PeerForwarder(final PluginSetting pluginSetting,
                          final PeerClientPool peerClientPool,
@@ -38,6 +43,7 @@ public class PeerForwarder extends AbstractPrepper<Record<ExportTraceServiceRequ
         this.peerClientPool = peerClientPool;
         this.hashRing = hashRing;
         this.maxNumSpansPerRequest = maxNumSpansPerRequest;
+        this.forwardRequestErrorCounter = pluginMetrics.counter(FORWARD_REQUEST_ERRORS);
     }
 
     public PeerForwarder(final PluginSetting pluginSetting) {
@@ -116,6 +122,7 @@ public class PeerForwarder extends AbstractPrepper<Record<ExportTraceServiceRequ
                 client.export(request);
             } catch (Exception e) {
                 LOG.error(String.format("Failed to forward the request:\n%s\n", request.toString()));
+                forwardRequestErrorCounter.increment();
                 localBuffer.add(new Record<>(request));
             }
         } else {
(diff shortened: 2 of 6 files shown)


wrijeff (commented):
Just as a general comment regarding the unit tests:

I noticed you're mocking the static Armeria client methods which I don't expect the PeerForwarder (and its test) should have to know about.

Check out this [PeerForwarder constructor](https://github.com/opendistro-for-elasticsearch/Data-Prepper/blob/master/data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java#L33-L35) which allows you to pass in mock objects directly, instead of having to configure PluginSetting perfectly right in each test. From there you can specify the mock PeerClientPool to return a mocked client, which you can then make assertions on.

I added it a few commits ago because I was running into this exact same problem. The way in which I chained the constructors together is a bit hacky, but it lets us invert the dependencies so that testing should be easier.

chenqi0805:
> Check out this PeerForwarder constructor which allows you to pass in mock objects directly, instead of having to configure PluginSetting perfectly right in each test. From there you can specify the mock PeerClientPool to return a mocked client, which you can then make assertions on.

Great call! Thanks for letting me know your thoughts put into this.

kowshikn:
Is this is only metric we plan to add in the Peer Forwarder?

chenqi0805:
> Is this is only metric we plan to add in the Peer Forwarder?

It is the only one I can think of for now other than the common ones (recordsIn, recordsOut, timeElapsed). Do you have other useful ones in mind?

kowshikn:
> > Is this is only metric we plan to add in the Peer Forwarder?
> 
> It is the only one I can think of for now other than the common ones (recordsIn, recordsOut, timeElapsed). Do you have other useful ones in mind?

Adding metrics should be done as a Operational person. If you are running this in production, what all things you would need to quickly identify issues in this plugin. I can think of,

1. Number of active peers -  This should help us to change in peers and when it happened.
2. Records Out per peer 
3. Time to send a request per peer
4. Failures per peer -  This will be vital, the peers will throw buffer full, we should probably wait and retry for ever.

Have a chat with @wrijeff and try to come up with metric we can drill down. 

P.S: I am yet to look at the latest version of this Plugin, will do it before the next revision.

chenqi0805:
@wrijeff @kowshikn @AustinTag As followup of discussion, I did a little summary and decided to first introduce the following set of custom metrics in peer forwarder:

- [ ] Timer
   - [ ] Timer on forwarding each request per target peer IP
- [ ] Counter
   - [ ] Counter on number of forwarded requests per target peer IP
   - [ ] Counter on number of forward request errors per target peer IP
- [ ] Gauge
   - [ ] Number of active peer IPs returned by PeerProviderList
 
# of requests ingested locally is already covered by recordsIn and recordsOut. So I tend to avoid redundant metrics. Let me know if I missed any outstanding important metrics.

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
```diff
@@ -24,20 +27,36 @@
...
+                         final PeerListProvider peerListProvider,
                          final PeerClientPool peerClientPool,
                          final HashRing hashRing,
                          final int maxNumSpansPerRequest) {
         super(pluginSetting);
         this.peerClientPool = peerClientPool;
         this.hashRing = hashRing;
         this.maxNumSpansPerRequest = maxNumSpansPerRequest;
+        this.peerListProvider = pluginMetrics.gauge(
+                PEER_ENDPOINTS, peerListProvider, provider -> (double) provider.getPeerList().size());
```
AustinTag, line 56: Nice, do we know how this works? How/when is the gauge value updated?

AustinTag (approved):
LGTM, just one question added as a comment

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
```diff
@@ -1,43 +1,46 @@
...
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.internal.verification.VerificationModeFactory.times;
-import static org.powermock.api.mockito.PowerMockito.doAnswer;
-import static org.powermock.api.mockito.PowerMockito.spy;
-
-@RunWith(PowerMockRunner.class)
```
wrijeff, line 37: I think these are the last references to PowerMock, should be able to remove it from the build.gradle file

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
wrijeff: Could this be an integer instead of double?

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
```diff
@@ -112,10 +132,19 @@ private void processRequest(final TraceServiceGrpc.TraceServiceBlockingStub clie
...
+            final Timer forwardRequestTimer = forwardRequestTimers.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.timer(String.format("%s:%s", FORWARD_REQUEST_LATENCY_PREFIX, ip)));
+            final Counter forwardedRequestCounter = forwardedRequestCounters.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_SUCCESS_PREFIX, ip)));
+            final Counter forwardRequestErrorCounter = forwardRequestErrorCounters.computeIfAbsent(
+                    peerIp, ip -> pluginMetrics.counter(String.format("%s:%s", FORWARD_REQUEST_ERRORS_PREFIX, ip)));
             try {
-                client.export(request);
+                forwardRequestTimer.record(() -> client.export(request));
+                forwardedRequestCounter.increment();
```
wrijeff, line 144: I'd be interested in knowing if we actually need both of these metrics. Could we potentially just count the number of timing metrics to derive the count (on the frontend UI, not in code).

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
```diff
@@ -95,45 +115,34 @@ public void testLocalIpOnly() {
         Assert.assertTrue(expectedResourceSpans.containsAll(exportedResourceSpans));
     }
 
-    @Test
-    public void testProcessRequest() throws Exception {
-        final PeerForwarder testPeerForwarder = generatePeerForwarder(Collections.singletonList(LOCAL_IP), 2);
```
wrijeff, line 100: Nice cleanup in this file 👍

data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
```diff
@@ -2,6 +2,8 @@
 
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import java.util.StringJoiner;
+import java.util.function.ToDoubleFunction;
+
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Metrics;
```
kowshikn, line 9: nit: Add a java doc on this class and link it micrometer docs. (Apologies for mentioning this in this PR, missed it the original PR that introduced this class)

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
chenqi0805: So the passed-in function will be invoked on the passed-in instance everytime Measurement.getValue is called, e.g.

```
@Test
    public void testUpdate() {
        Foo foo = new Foo();

        MetricsTestUtil.initMetrics();

        PLUGIN_METRICS.gauge("gauge", foo, obj -> obj.getSize());

        List<Measurement> res = MetricsTestUtil.getMeasurementList(new StringJoiner(MetricNames.DELIMITER)
                .add(PIPELINE_NAME).add(PLUGIN_NAME)
                .add("gauge").toString());
        Assert.assertEquals(0.0, res.get(0).getValue(), 0);

        foo.a.add("sth");
        Assert.assertEquals(1.0, res.get(0).getValue(), 0);
    }

    static class Foo {
        List<String> a = new ArrayList<>();

        public int getSize() {
            return a.size();
        }
    }
```
I was not able to mock PeerListProviderFactory::new in the context of PeerForwarder so there is no test case here to manifest that.

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
```diff
@@ -57,6 +60,23 @@
...
+    private MockedStatic<PeerClientPool> peerClientPoolClassMock;
+
+    @Mock
+    private PeerClientPool peerClientPool;
+
+    @Before
+    public void setUp() {
+        peerClientPoolClassMock = Mockito.mockStatic(PeerClientPool.class);
+        peerClientPoolClassMock.when(PeerClientPool::getInstance).thenReturn(peerClientPool);
+    }
```
wrijeff, line 72: One way to get around having to mock static here would be using the multi-parameter constructor to build the test object, something like:
```
    @Before
    public void setup() {
        peerForwarder =  new PeerForwarder(mockPluginSetting,
                mockPeerForwarder,
                mockHashing,
                maxRequestsValue);
    }
```
This keeps the unit test from having to know how to construct a PeerForwarder (that functionality is currently being handled/tested in PeerForwarderConfig and PeerForwarderConfigTest).

Doesn't have to change in this PR, just a thought (also my opinion, I'm sure some would prefer to test actual objects when possible but there are pros/cons to each).

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
kowshikn, line 56: I think we spoke about this in the slack channel, is it hard to provide the `pluginMetrics` as a reference to PeerListProviderFactory? So that both static and dynamic can emit metrics on their own like errors, refresh etc.

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
```diff
@@ -95,45 +115,34 @@ public void testLocalIpOnly() {
...
-        Assert.assertEquals(1, localBuffer.size());
-        verify(mockClient, times(1)).export(REQUEST_2);
-    }
-
     @Test
     public void testSingleRemoteIpBothLocalAndForwardedRequest() throws Exception {
         final List<String> testIps = generateTestIps(2);
-        final PeerForwarder testPeerForwarder = spy(generatePeerForwarder(testIps, 3));
+        final Channel channel = mock(Channel.class);
+        final TraceServiceGrpc.TraceServiceBlockingStub client = mock(TraceServiceGrpc.TraceServiceBlockingStub.class);
```
wrijeff, line 122: For mock objects consistent across multiple tests, could factor out by declaring as instances variables like:
```
@Mock
private Channel channel;
...
@Mock 
private TraceServiceGrpc.TraceServiceBlockingStub client;
```
The `@RunWith(MockitoJUnitRunner.class)` you have at the top will automatically call `initMocks()` and create mock objects where it sees the `@Mock` annotation.

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
kowshikn: > Nice, do we know how this works? How/when is the gauge value updated?

Gauge has a weak-reference and auto-updates.

kowshikn:
Discussed offline with @wrijeff  and @chenqi0805, we will have the refactor of passing the `pluginMetrics` in a separate PR.

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
chenqi0805: > Could this be an integer instead of double?

Actually the cast to double might not be necessary. But in test case `.getValue()` still returns double.

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
chenqi0805: Good question! So in the results of timer, there is indeed a TOTAL number of calls. The issue is that it will count # of calls on client.export no matter it fails or not. Therefore we still need independent counters for success and failures.

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
chenqi0805: > This keeps the unit test from having to know how to construct a PeerForwarder (that functionality is currently being handled/tested in PeerForwarderConfig and PeerForwarderConfigTest).

Agreed on that the design will make testing config and main class more independent, but I would assume here PeerForwarderConfig should work fine. Otherwise it should be captured in PeerForwarderConfigTest.

Also, I am not sure how it relates to avoiding PeerClientPool static mock since it is still a singleton we want to reset for each test case.

data-prepper-plugins/peer-forwarder/src/test/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarderTest.java
chenqi0805: Good catch. Should also globalize client instance

data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/PluginMetrics.java
chenqi0805: Sure.

wrijeff (approved):

data-prepper-plugins/peer-forwarder/src/main/java/com/amazon/dataprepper/plugins/processor/peerforwarder/PeerForwarder.java
chenqi0805: ref #257

PR #273, opened 2021-01-06
ENH: increase coverage on elasticsearch sink

chenqi0805:
*Issue #, if available:*
#239 

*Description of changes:*
- Added test cases to increase coverage on elasticsearch sink
- Enforce 90% coverage on each class except for the ones that are covered by ESRestTestCase

Note: Have not yet found a way to include integTest into jacoco.


data-prepper-plugins/elasticsearch/build.gradle
```diff
@@ -84,6 +84,34 @@ integTest.runner {
     systemProperty "password", System.getProperty("password")
 }
 
+jacocoTestReport {
+    dependsOn test
+    afterEvaluate {
+        classDirectories.from = files(classDirectories.files.collect {
+            fileTree(dir: it, exclude: [
+                    'com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.class',
```
AustinTag, line 92: Is this excluding the ElasticsearchSink class from the code coverage?

kowshikn (approved):

data-prepper-plugins/elasticsearch/build.gradle
chenqi0805: Yes. It is supposed to be covered by ElasticsearchSinkIT under the elasticsearch integration test framework. I had tried to apply jacoco to that but unfortunately it seemed not to produce meaningful results as I expect. Right now this class is 0% coverage if we do not exclude.

AustinTag: I see. Are we ok with that? Is the class not unit-testable?

chenqi0805: The path I can think of to cover it by unit test is to mock out the sdk of ES rest client. This will certainly produce good percentage of the coverage. But I am not sure about the quality of the number if we cover it this way. I can add a TODO in the build.gradle to note on include those classes into coverage.

AustinTag: Sure, that's fine. We should strive to not exclude classes unless they really aren't testable or are not used.

AustinTag (approved):

PR #274, opened 2021-01-07
Adding k8s example files under the /dev directory.

wrijeff:
*Description of changes:*
* Added example Kubernetes config files to run the trace sample app locally with k8s


chenqi0805 (approved):
LGTM. Ran successfully as a user. 😃

wrijeff:
> LGTM. Ran successfully as a user. 😃

Nice!

kowshikn:
As discussed offline, K8s best practice is to to keep things in a single yaml. I would like us to follow it and keep all the sample project dependencies like Sample App, OTelCollector, ODFE in one file and have one yaml file for Data Prepper. This will make it more readable. 

Also, we should create the release story for readme which is more focused on how to run Data Prepper in static and K8 mode. That will be more useful.

wrijeff:
Pushed a change to build locally, also enabled PeerForwarder in the /dev configs (plus a few other misc cleanups).

Got distracted because I wasn't able to see PeerForwarder logs, at first I thought there might have been a regression but tracked the issue down to be purely logging-related. The plugin works, but log thresholds are misconfigured somewhere (log.error appears correctly but log.info does not).

Will follow up on sample project logging in a separate effort.

kowshikn (approved):

PR #279, opened 2021-01-13
First cut of adding metrics to Service Map Stateful Prepper

sshivanii:
*Issue #, if available:*

*Description of changes:*

- Initial cut of adding the following metrics to the Service Map Stateful Prepper:

  - 1. uniqueRelationshipsCounter
  - 2. uniqueTraceGroupsCounter

### DIFF
--- a/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
+++ b/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
@@ -9,6 +9,7 @@ import com.amazon.dataprepper.plugins.prepper.state.MapDbPrepperState;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.primitives.SignedBytes;
+import io.micrometer.core.instrument.Counter;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -120,6 +135,7 @@ private static File createPath(File path) {
                                         if (span.getParentSpanId().isEmpty()) {
                                             try {
                                                 currentTraceGroupWindow.put(span.getTraceId().toByteArray(), span.getName());
+                                                uniqueTraceGroupsCounter.increment();
```
AustinTag, line 138: This is not unique trace groups. This number will be equal to the unique number of traces

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -154,6 +170,7 @@ private static File createPath(File path) {
                             .map(serviceMapRelationship -> {
                                 try {
                                     relationshipState.add(serviceMapRelationship);
+                                    uniqueRelationshipsCounter.increment();
```
AustinTag, line 173: This may be a little confusing for the end user, as this number will likely be much higher than the number of relationships they see in the service map plugin. We may need to think a bit more about how we want to calculate this.

AustinTag (commented):
In general, we should not add new logic like this without adding corresponding tests.

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
sshivanii: Looking into it

sshivanii:
Had an offline discussion with @AustinTag and we want different set of metrics, not these. Hence, closing this PR.

Issue #201, opened 2020-12-08
Add monitoring support for Data Prepper


kowshikn:
Data Prepper should provide a way for users to monitor the various components that are running. We will expose a GET API endpoint which allows the user to get the current state of metrics for the running Data Prepper. 

While certain components have specific metrics that apply to them, there are also certain metrics which apply more generically to processors, sinks, and buffers. The following metrics will be provided by default to all components of that type:

- Prepper: Records in, records out, time elapsed
- Sink: Records out, time elapsed
- Buffer: Records written, records read, write time elapsed, read time elapsed, and timeouts

For implementation, we will use Micrometer.io for metrics instrumentation, and Prometheus for the metrics backend. Micrometer allows us to decouple metrics instrumentation from backend implementation, and swap metrics backends if needed. Prometheus is widely used and supported, and provides customers with many options for consumption.

AustinTag:
Data Prepper should provide a way for users to monitor the various components that are running. We will expose a GET API endpoint which allows the user to get the current state of metrics for the running Data Prepper. 

While certain components have specific metrics that apply to them, there are also certain metrics which apply more generically to processors, sinks, sources, and buffers. The following metrics will be provided by default to all components of that type:

- Prepper: Records in, records out, time elapsed
- Sink: Records out, time elapsed
- Buffer: Records written, records read, write time elapsed, read time elapsed, and timeouts
- Source: Records out

For implementation, we will use Micrometer.io for metrics instrumentation, and Prometheus for the metrics backend. Micrometer allows us to decouple metrics instrumentation from backend implementation, and swap metrics backends if needed. Prometheus is widely used and supported, and provides customers with many options for consumption.

kowshikn:
@chenqi0805 Can you check with Austin and close this issue if its complete.

AustinTag:
The only thing left on this issue is to normalize naming across metrics. For example: 

span_errors_counter
bulk_request_errors

These are both error counters but have different naming patterns. This should be a quick fix and then we can close out this task.

chenqi0805:
@AustinTag I doublechecked current master and it is all camel case now. Or do you mean we should give them the same name? AFAICT, the former is still different from the later as a bulkRequest will include multiple spans.

AustinTag:
I see. I'm not concerned with the naming of objects in the code so much as the metric name that is exposed externally. If all of those follow the same pattern then I think we are good to close this issue.

PR #281, opened 2021-01-19
First pass of adding metrics to Service Map Stateful Prepper

sshivanii:
*Issue #, if available:*
#201 

*Description of changes:*

- Adds sizeInBytes function to MapDbPrepper state to retrieve file size in bytes
- Adds sizeInBytes function to interface PrepperState
- Test case to check sizeInBytes

TODO:

- [x] Implementation of gauges spansDbSize and traceGroupDbSize in ServiceMapStatefulPrepper
- [x] Test cases to verify the gauges

### DIFF
--- a/data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/prepper/state/PrepperStateTest.java
+++ b/data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/prepper/state/PrepperStateTest.java
@@ -33,7 +33,6 @@ public abstract class PrepperStateTest {
         prepperState.put(key2, data2);
 
         Assert.assertEquals(2, prepperState.size());
-        Assert.assertEquals(2097152, prepperState.sizeInBytes());
     }
 
     @Test
--- a/data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/prepper/state/PrepperStateTest.java
+++ b/data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/prepper/state/PrepperStateTest.java
@@ -33,6 +33,7 @@ public abstract class PrepperStateTest {
         prepperState.put(key2, data2);
 
         Assert.assertEquals(2, prepperState.size());
+        Assert.assertEquals(2097152, prepperState.sizeInBytes());
     }
 
     @Test
--- a/data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
+++ b/data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
@@ -154,7 +154,7 @@ public class LmdbPrepperState<T> implements PrepperState<byte[], T> {
      */
     @Override
     public long sizeInBytes() {
-        return 2097152;
+        return 0;
     }
 
     private KeyRange getRange(int segments, int index) {
--- a/data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
+++ b/data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
@@ -28,6 +28,7 @@ public class LmdbPrepperState<T> implements PrepperState<byte[], T> {
     private final Class<T> clazz; //Needed for deserialization
     private final File dbFile;
 
+
     /**
      * Constructor for LMDB prepper state. See LMDB-Java for more info:
      * https://github.com/lmdbjava/lmdbjava
@@ -145,6 +146,15 @@ public class LmdbPrepperState<T> implements PrepperState<byte[], T> {
         try (Txn<ByteBuffer> txn = env.txnRead()) {
             return db.stat(txn).entries;
         }
+    }
+
+    /**
+     * TODO: Implement the sizeInBytes function
+     * @return Size of the prepper state file, in bytes.
+     */
+    @Override
+    public long sizeInBytes() {
+        return 2097152;
     }
 
     private KeyRange getRange(int segments, int index) {
--- a/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
+++ b/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
@@ -37,12 +37,14 @@ public class MapDbPrepperState<V> implements PrepperState<byte[], V> {
                         .fileMmapEnable() //MapDB uses the (slower) Random Access Files by default
                         .fileMmapPreclearDisable()
                         .executorEnable()
+                        .transactionEnable()
+                        .closeOnJvmShutdown()
                         .concurrencyScale(concurrencyScale)
                         .make()
                         .treeMap(dbName)
                         .counterEnable() //Treemap doesnt keep:q size counter by default
                         .keySerializer(SIGNED_BYTE_ARRAY_SERIALIZER)
-                        .valueSerializer(Serializer.JAVA).create();
+                        .valueSerializer(Serializer.JAVA).createOrOpen();
     }
 
     @Override
--- a/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
+++ b/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
@@ -27,10 +27,12 @@ public class MapDbPrepperState<V> implements PrepperState<byte[], V> {
     private static final SignedByteArraySerializer SIGNED_BYTE_ARRAY_SERIALIZER = new SignedByteArraySerializer();
 
     private final BTreeMap<byte[], V> map;
+    private final File completeDbPath;
 
     public MapDbPrepperState(final File dbPath, final String dbName, final int concurrencyScale) {
+        this.completeDbPath = new File(String.join("/", dbPath.getPath(), dbName));
         map =
-                (BTreeMap<byte[], V>) DBMaker.fileDB(new File(String.join("/", dbPath.getPath(), dbName)))
+                (BTreeMap<byte[], V>) DBMaker.fileDB(completeDbPath)
                         .fileDeleteAfterClose()
                         .fileMmapEnable() //MapDB uses the (slower) Random Access Files by default
                         .fileMmapPreclearDisable()
@@ -106,6 +108,12 @@ public class MapDbPrepperState<V> implements PrepperState<byte[], V> {
     @Override
     public long size() {
         return map.size();
+    }
+
+
+    @Override
+    public long sizeInBytes() {
+        return completeDbPath.length();
     }
 
     @Override
--- a/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
+++ b/data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
@@ -27,12 +27,12 @@ public class MapDbPrepperState<V> implements PrepperState<byte[], V> {
     private static final SignedByteArraySerializer SIGNED_BYTE_ARRAY_SERIALIZER = new SignedByteArraySerializer();
 
     private final BTreeMap<byte[], V> map;
-    private final File completeDbPath;
+    private final File dbFile;
 
     public MapDbPrepperState(final File dbPath, final String dbName, final int concurrencyScale) {
-        this.completeDbPath = new File(String.join("/", dbPath.getPath(), dbName));
+        this.dbFile = new File(String.join("/", dbPath.getPath(), dbName));
         map =
-                (BTreeMap<byte[], V>) DBMaker.fileDB(completeDbPath)
+                (BTreeMap<byte[], V>) DBMaker.fileDB(dbFile)
                         .fileDeleteAfterClose()
                         .fileMmapEnable() //MapDB uses the (slower) Random Access Files by default
                         .fileMmapPreclearDisable()
@@ -113,7 +113,7 @@ public class MapDbPrepperState<V> implements PrepperState<byte[], V> {
 
     @Override
     public long sizeInBytes() {
-        return completeDbPath.length();
+        return dbFile.length();
     }
 
     @Override
--- a/data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
+++ b/data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
@@ -62,7 +62,7 @@ public class MapDbPrepperStateTest extends PrepperStateTest {
                 data4.stringVal
         )));
 
-        Assert.assertEquals(2097152, prepperState.sizeInBytes());
+        Assert.assertTrue(prepperState.sizeInBytes() > 0);
 
 
     }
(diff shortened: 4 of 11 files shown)


data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
```diff
@@ -27,10 +27,12 @@ public int compare(byte[] o1, byte[] o2) {
     private static final SignedByteArraySerializer SIGNED_BYTE_ARRAY_SERIALIZER = new SignedByteArraySerializer();
 
     private final BTreeMap<byte[], V> map;
+    private final File completeDbPath;
```
AustinTag, line 30: Nit: Why not just call this the dbFile? This is a file, not a path.

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -9,14 +9,20 @@
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.primitives.SignedBytes;
+import io.micrometer.core.instrument.Counter;
```
AustinTag, line 12: This stuff needs to all be removed.

data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
```diff
@@ -147,6 +148,15 @@ public long size() {
         }
     }
 
+    /**
+     * TODO: Implement the sizeInBytes function
+     * @return Size of the prepper state file, in bytes.
+     */
+    @Override
+    public long sizeInBytes() {
+        return 2097152;
```
AustinTag, line 157: This is not really right. Lets move the db size test to each individual prepper state class test, because those will be different between the implementations. For now, you can return 0 from this.

data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/prepper/state/PrepperStateTest.java
```diff
@@ -33,6 +33,7 @@ public void testSize() {
         prepperState.put(key2, data2);
 
         Assert.assertEquals(2, prepperState.size());
+        Assert.assertEquals(2097152, prepperState.sizeInBytes());
```
AustinTag, line 36: This will not work when you implement the size function for lmdb, because sizes will not be the same. For example, MapDB allocates file size in MB increments, which is why this is exactly 2MB.

AustinTag (changes requested):

data-prepper-plugins/lmdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/LmdbPrepperState.java
sshivanii: Will do

data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
```diff
@@ -27,20 +27,24 @@ public int compare(byte[] o1, byte[] o2) {
...
                         .executorEnable()
+                        .transactionEnable()
+                        .closeOnJvmShutdown()
                         .concurrencyScale(concurrencyScale)
                         .make()
                         .treeMap(dbName)
                         .counterEnable() //Treemap doesnt keep:q size counter by default
                         .keySerializer(SIGNED_BYTE_ARRAY_SERIALIZER)
-                        .valueSerializer(Serializer.JAVA).create();
+                        .valueSerializer(Serializer.JAVA).createOrOpen();
```
AustinTag, line 47: Why do we need this line? Is it because we already created the dbFile on line 33?

data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
```diff
@@ -62,6 +62,9 @@ public String apply(byte[] bytes, DataClass s) {
                 data4.stringVal
         )));
 
+        Assert.assertTrue(prepperState.sizeInBytes() > 0);
```
AustinTag, line 65: Is this the best we can do here? It should be clear what the size is right? I thought a previous revision had an exact size

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -41,17 +51,20 @@
...
     //TODO: Consider keeping this state in lmdb
-    private volatile static  HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
+    private volatile static HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
     private static File dbPath;
     private static Clock clock;
     private static int processWorkers;
 
     private final int thisPrepperId;
+    private final Pair<MapDbPrepperState<ServiceMapStateData>, MapDbPrepperState<ServiceMapStateData>> spanDbPair;
+    private final Pair<MapDbPrepperState<String>, MapDbPrepperState<String>> traceDbPair;
```
AustinTag, line 61: Have you tried creating the pair inside of the call to plugin metrics? I am not sure if the object will get garbage collected, but it would be better to not have to declare these pairs if possible.

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -10,13 +10,19 @@
...
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.Serializable;
 import java.time.Clock;
-import java.util.*;
+import java.util.Collection;
+import java.util.Collections;
```
AustinTag, line 21: might be some extra imports up here

data-prepper-plugins/service-map-stateful/src/test/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepperTest.java
```diff
@@ -172,21 +186,33 @@ public void testTraceGroups() throws Exception {
 
         Assert.assertTrue(evaluateEdges(relationshipsFound).containsAll(expectedSourceDests));
 
+        // Verify gauges
+        final List<Measurement> spansDbSizeMeasurement = MetricsTestUtil.getMeasurementList(
+                new StringJoiner(MetricNames.DELIMITER).add("testPipelineName").add("testServiceMapPrepper")
+                        .add(ServiceMapStatefulPrepper.SPANS_DB_SIZE).toString());
+        Assert.assertEquals(1, spansDbSizeMeasurement.size());
```
AustinTag, line 193: I think you're not verifying the value of the metrics here right?

data-prepper-plugins/service-map-stateful/build.gradle
```diff
@@ -16,10 +16,13 @@ dependencies {
...
     compile project(':data-prepper-plugins:lmdb-prepper-state')
     compile project(':data-prepper-plugins:mapdb-prepper-state')
+    compile 'org.apache.commons:commons-lang3:3.11'
+    testCompile project(':data-prepper-api').sourceSets.test.output
+    implementation "io.micrometer:micrometer-core:1.5.1"
     implementation "com.fasterxml.jackson.core:jackson-databind:${versionMap.jackson_databind}"
     implementation "io.opentelemetry:opentelemetry-proto:${versionMap.opentelemetry_proto}"
     testImplementation "org.hamcrest:hamcrest:${versionMap.hamcrest}"
-    testImplementation "org.mockito:mockito-core:${versionMap.mockito}"
+    testImplementation "org.mockito:mockito-inline:3.6.28"
```
AustinTag, line 25: any reason for this change here?

data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
sshivanii: Yes, this change is to ensure that if we have already created the Db, we open the existing one else it results in an error saying the named db already exists.

data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
sshivanii: I changed this in case the size is varied. I can revert this.

data-prepper-plugins/service-map-stateful/build.gradle
sshivanii: Keeping it consistent with the Mockito dependency in other plugins. The Otel trace source already has a TODO in the gradle build file to update the version map to a higher version of mockito in all subprojects

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
sshivanii: Both the Collection and Collections imports are needed in this class. Since we don't want any wildcard imports in our project, I configured my IntelliJ so it explicitly added these single exports.

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
sshivanii: I will look into this.

data-prepper-plugins/service-map-stateful/src/test/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepperTest.java
sshivanii: Oh right, need to add that.

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
AustinTag: I see, you replaced the wildcard with the individual imports.

data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
AustinTag: Well if the size varies, we should just verify that its in the range in which is varies. > 0 is probably not restrictive enough to catch issues if they arise.

data-prepper-plugins/mapdb-prepper-state/src/main/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperState.java
AustinTag: So nothing about your code makes it so that the db should have already been created right?

sshivanii: That's right

data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
sshivanii: Makes sense, will do

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -10,13 +10,19 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.primitives.SignedBytes;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
+import org.apache.commons.lang3.tuple.Pair;
```
AustinTag, line 13: not needed

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -41,17 +50,20 @@
...
     private volatile static MapDbPrepperState<String> currentTraceGroupWindow;
     //TODO: Consider keeping this state in lmdb
-    private volatile static  HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
+    private volatile static HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
     private static File dbPath;
     private static Clock clock;
     private static int processWorkers;
 
     private final int thisPrepperId;
+    private Pair<MapDbPrepperState<ServiceMapStateData>, MapDbPrepperState<ServiceMapStateData>> spanDbPair;
```
AustinTag, line 59: not needed

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -72,17 +85,23 @@ public ServiceMapStatefulPrepper(final long windowDurationMillis,
             currentTraceGroupWindow = new MapDbPrepperState<>(dbPath, getNewTraceDbName(), processWorkers);
             previousTraceGroupWindow = new MapDbPrepperState<>(dbPath, getNewTraceDbName() + EMPTY_SUFFIX, processWorkers);
         }
+
+        this.spanDbPair = Pair.of(previousWindow, currentWindow);
```
AustinTag, line 89: not needed

data-prepper-plugins/mapdb-prepper-state/src/test/java/com/amazon/dataprepper/plugins/prepper/state/MapDbPrepperStateTest.java
```diff
@@ -62,6 +62,7 @@ public String apply(byte[] bytes, DataClass s) {
                 data4.stringVal
         )));
 
+        Assert.assertTrue( prepperState.sizeInBytes() > 0 && prepperState.sizeInBytes() <= 2097152);
```
AustinTag, line 65: Do we need to use 0 < size < 2097152 ? This should be a static size i think

AustinTag (changes requested):

AustinTag (approved):

data-prepper-plugins/service-map-stateful/build.gradle
```diff
@@ -16,10 +16,12 @@ dependencies {
     compile project(':data-prepper-plugins:common')
     compile project(':data-prepper-plugins:lmdb-prepper-state')
     compile project(':data-prepper-plugins:mapdb-prepper-state')
+    testCompile project(':data-prepper-api').sourceSets.test.output
+    implementation "io.micrometer:micrometer-core:1.5.1"
     implementation "com.fasterxml.jackson.core:jackson-databind:${versionMap.jackson_databind}"
     implementation "io.opentelemetry:opentelemetry-proto:${versionMap.opentelemetry_proto}"
     testImplementation "org.hamcrest:hamcrest:${versionMap.hamcrest}"
-    testImplementation "org.mockito:mockito-core:${versionMap.mockito}"
+    testImplementation "org.mockito:mockito-inline:3.6.28"
```
wrijeff, line 24: Might be time to move mockito-inline to the version map (I see there are 3 other uses of it). If it's not trivial can be a separate effort.

wrijeff (approved):

Issue #285, opened 2021-01-25
Add max connection count and thread count configuration options to OTelTraceSource


wrijeff:
**Is your feature request related to a problem? Please describe.**
Enhancement - the OTelTraceSource gRPC request-handling server has default values for max connections and worker threads. Expose these settings via config.

PR #286, opened 2021-01-25
#285: Added max connection count and thread count configuration optio…

wrijeff:
…ns to OTelTraceSource.

*Issue #, if available:* #285

*Description of changes:*
* Setting default max connections to be 500 (down from Armeria default of Integer.MAX_VALUE)
* Setting default max threads to be 200 (same as Armeria's default)

Armeria docs: https://armeria.dev/docs/advanced-production-checklist/. We'll want to make sure these are sane values when we start coming up with instance sizing guidelines, but I think these should be decent for now.

### DIFF
--- a/examples/dev/trace-analytics-sample-app/docker-compose.yml
+++ b/examples/dev/trace-analytics-sample-app/docker-compose.yml
@@ -62,8 +62,8 @@ services:
       - ../../../shared-config/log4j.properties:/usr/share/data-prepper/log4j.properties
       - ../../demo/root-ca.pem:/usr/share/data-prepper/root-ca.pem
     ports:
-      - "4900:4900"
-      - "21890:21890"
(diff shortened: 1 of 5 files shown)


AustinTag (approved):
LGTM 👍

examples/dev/trace-analytics-sample-app/docker-compose.yml
```diff
@@ -62,6 +62,7 @@ services:
       - ../../../shared-config/log4j.properties:/usr/share/data-prepper/log4j.properties
       - ../../demo/root-ca.pem:/usr/share/data-prepper/root-ca.pem
     ports:
+      - "4900:4900"
```
chenqi0805, line 65: Might better add a comment that this is the data prepper server port.

chenqi0805 (approved):
Just a nit. LGTM!

Issue #216, opened 2020-12-10
Checkpoint interface in Buffer plugin


kowshikn:
The default buffer of the Data Prepper is a `blocking queue`. The blocking queue has `buffer_size` which is the capacity of the buffer. 

*Concern*

When the processors read from the buffer the records in the buffer become inflight and we accept more requests. This is brings in a situation where the inflight requests are more than the actual buffer size. In general we should make sure Data Prepper doesn't have more records than the buffer capacity.  

*Ideas*

We should introduce the interface `checkpoint()` and have counter based implementation for the `blocking queue` variant.

kowshikn:
@erosas @yadavcbala feel free to add your thoughts.

kowshikn:
This is complete we have the checkpoint state in the buffer.

PR #287, opened 2021-01-25
ENH: initial checkpoint model and API

chenqi0805:
*Issue #, if available:*
#216 

*Description of changes:*
This PR 

1. Draft the initial model of CheckpointStates that takes in the number of records to be checked.
2. Add checkpoint API into Buffer interface and fix relevant places
3. Invoke the checkpoint API in the processWorker.
4. Modified BlockingBuffer to add number of inflight records as state variable.
5. Implemented checkpoint API in BlockingBuffer to keep track of inflight data
6. Modified doWrite API in BlockingBuffer to control inflight data below buffer capacity by introducing lock with explicit condition variable.

Note: Test cases that cover BlockingBuffer are synchronized. I expect async features to be at least partially covered by e2e tests. But if necessary, we should try to device those async tests.

### DIFF
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
@@ -40,6 +40,7 @@ public class ProcessWorker implements Runnable {
         try {
             do {
                 Collection records = readBuffer.read(pipeline.getReadBatchTimeoutInMillis());
+                final CheckpointState checkpointState = new CheckpointState(records.size());
                 //TODO Hacky way to avoid logging continuously - Will be removed as part of metrics implementation
                 if (records.isEmpty()) {
                     if(!isEmptyRecordsLogged) {
@@ -55,8 +56,9 @@ public class ProcessWorker implements Runnable {
                 }
                 if (!records.isEmpty()) {
                     postToSink(records); //TODO use the response to ack the buffer on failure?
-                    readBuffer.checkpoint(new CheckpointState(records.size()));
                 }
+                // Checkpoint the current batch read from the buffer after being processed by prepper and sinks.
+                readBuffer.checkpoint(checkpointState);
             } while (!shouldStop());
         } catch (final Exception ex) {
             LOG.error("Encountered exception during pipeline processing", ex); //do not halt the execution
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
@@ -6,6 +6,7 @@ import com.amazon.dataprepper.model.record.Record;
 import com.amazon.dataprepper.model.sink.Sink;
 import com.amazon.dataprepper.pipeline.common.FutureHelper;
 import com.amazon.dataprepper.pipeline.common.FutureHelperResult;
+import com.amazon.dataprepper.model.CheckpointState;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -54,6 +55,7 @@ public class ProcessWorker implements Runnable {
                 }
                 if (!records.isEmpty()) {
                     postToSink(records); //TODO use the response to ack the buffer on failure?
+                    readBuffer.checkpoint(new CheckpointState(records.size()));
                 }
             } while (!shouldStop());
         } catch (final Exception ex) {
--- a/data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
+++ b/data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
@@ -6,6 +6,7 @@ import com.amazon.dataprepper.model.buffer.AbstractBuffer;
 import com.amazon.dataprepper.model.buffer.Buffer;
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.record.Record;
+import com.amazon.dataprepper.model.CheckpointState;
 import com.google.common.base.Stopwatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -19,6 +20,9 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.lang.String.format;
@@ -41,8 +45,15 @@ public class BlockingBuffer<T extends Record<?>> extends AbstractBuffer<T> {
     private static final String ATTRIBUTE_BATCH_SIZE = "batch_size";
 
     private final int batchSize;
+    private final int bufferCapacity;
     private final BlockingQueue<T> blockingQueue;
     private final String pipelineName;
+    private final AtomicInteger numInflightRecords = new AtomicInteger();
+
+    /** Lock held by doWrite */
+    private final ReentrantLock writeLock = new ReentrantLock();
+    /** Wait queue for waiting doWrites. Needs to be guarded by writeLock */
+    private final Condition notFull = writeLock.newCondition();
 
     /**
      * Creates a BlockingBuffer with the given (fixed) capacity.
@@ -54,6 +65,7 @@ public class BlockingBuffer<T extends Record<?>> extends AbstractBuffer<T> {
     public BlockingBuffer(final int bufferCapacity, final int batchSize, final String pipelineName) {
         super("BlockingBuffer", pipelineName);
         this.batchSize = batchSize;
+        this.bufferCapacity = bufferCapacity;
         this.blockingQueue = new LinkedBlockingQueue<>(bufferCapacity);
         this.pipelineName = pipelineName;
     }
@@ -80,15 +92,29 @@ public class BlockingBuffer<T extends Record<?>> extends AbstractBuffer<T> {
 
     @Override
     public void doWrite(T record, int timeoutInMillis) throws TimeoutException {
+        long nanos = TimeUnit.MILLISECONDS.toNanos(timeoutInMillis);
+        final int c;
         try {
-            boolean isSuccess = blockingQueue.offer(record, timeoutInMillis, TimeUnit.MILLISECONDS);
-            if (!isSuccess) {
-                throw new TimeoutException(format("Pipeline [%s] - Buffer is full, timed out waiting for a slot",
-                        pipelineName));
+            writeLock.lockInterruptibly();
+            final AtomicInteger count = this.numInflightRecords;
+            while (count.get() == this.bufferCapacity) {
+                if (nanos <= 0L) {
+                    throw new TimeoutException(format("Pipeline [%s] - Buffer is full, timed out waiting for a slot",
+                            pipelineName));
+                }
+                nanos = notFull.awaitNanos(nanos);
+            }
+            blockingQueue.offer(record);
+            c = count.incrementAndGet();
+            if (c < this.bufferCapacity) {
+                // Wake up other doWrite waiting threads due to previous full buffer state.
+                notFull.signal();
             }
         } catch (InterruptedException ex) {
             LOG.error("Pipeline [{}] - Buffer is full, interrupted while waiting to write the record", pipelineName, ex);
             throw new TimeoutException("Buffer is full, timed out waiting for a slot");
+        } finally {
+            writeLock.unlock();
         }
     }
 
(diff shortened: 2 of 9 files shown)


data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -80,15 +92,29 @@ public BlockingBuffer(final String pipelineName) {
 
     @Override
     public void doWrite(T record, int timeoutInMillis) throws TimeoutException {
+        long nanos = TimeUnit.MILLISECONDS.toNanos(timeoutInMillis);
+        final int c;
```
AustinTag, line 96: nit: why declare c here? if you're not using it in the catch/finally block, it can be declared inline with the initialization

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -80,15 +92,29 @@ public BlockingBuffer(final String pipelineName) {
...
         try {
-            boolean isSuccess = blockingQueue.offer(record, timeoutInMillis, TimeUnit.MILLISECONDS);
-            if (!isSuccess) {
-                throw new TimeoutException(format("Pipeline [%s] - Buffer is full, timed out waiting for a slot",
-                        pipelineName));
+            writeLock.lockInterruptibly();
```
AustinTag, line 98: maybe this line through 106 can go into a function that is named intuitively. this write function is becoming a little unreadable. something like "waitForWriteLockAndBufferCapacity()" (but maybe less wordy)

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -80,15 +92,29 @@ public BlockingBuffer(final String pipelineName) {
...
+                    throw new TimeoutException(format("Pipeline [%s] - Buffer is full, timed out waiting for a slot",
+                            pipelineName));
+                }
+                nanos = notFull.awaitNanos(nanos);
+            }
+            blockingQueue.offer(record);
+            c = count.incrementAndGet();
+            if (c < this.bufferCapacity) {
+                // Wake up other doWrite waiting threads due to previous full buffer state.
+                notFull.signal();
```
AustinTag, line 111: should this be in the finally?

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -132,4 +158,27 @@ public static PluginSetting getDefaultPluginSettings() {
         settings.put(ATTRIBUTE_BATCH_SIZE, DEFAULT_BATCH_SIZE);
         return new PluginSetting(PLUGIN_NAME, settings);
     }
+
+    @Override
+    public void checkpoint(final CheckpointState checkpointState) {
+        final int numCheckedRecords = checkpointState.getNumCheckedRecords();
+        final int c = numInflightRecords.getAndAdd(-numCheckedRecords);
```
AustinTag, line 165: should numInFlightRecords be updated here?

AustinTag: im not actually sure where numCheckedRecords is updated

data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/buffer/BlockingBufferTests.java
```diff
@@ -62,14 +64,35 @@ public void testInsertNull() throws TimeoutException {
...
-    public void testNoEmptySpace() throws TimeoutException {
+    public void testNoEmptySpaceWriteOnly() throws TimeoutException {
         final BlockingBuffer<Record<String>> blockingBuffer = new BlockingBuffer<>(1, TEST_BATCH_SIZE,
                 TEST_PIPELINE_NAME);
         assertThat(blockingBuffer, notNullValue());
         blockingBuffer.write(new Record<>("FILL_THE_BUFFER"), TEST_WRITE_TIMEOUT);
         blockingBuffer.write(new Record<>("TIMEOUT"), TEST_WRITE_TIMEOUT);
     }
 
+    @Test
```
AustinTag, line 75: These tests should have comments describing what they do, because they are not exactly readable. One possible format is the if-when-then format:

//Describe setup (If)

//Describe test actions (when)

//Describe assertions (then)

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -41,8 +45,15 @@
     private static final String ATTRIBUTE_BATCH_SIZE = "batch_size";
 
     private final int batchSize;
+    private final int bufferCapacity;
     private final BlockingQueue<T> blockingQueue;
     private final String pipelineName;
+    private final AtomicInteger numInflightRecords = new AtomicInteger();
```
AustinTag, line 51: These are not exactly "inflight" records. These are uncheckpointed records that have been written. In-flight would refer to records that have been read but not yet checkpointed. My understanding was that we did not want the in-flight to be greater than buffer capacity. But it seems that what we have here is in-flight + unread cannot be greater than buffer capacity. Just want to clarify that this is what we want.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
```diff
@@ -54,6 +55,7 @@ public void run() {
                 }
                 if (!records.isEmpty()) {
                     postToSink(records); //TODO use the response to ack the buffer on failure?
+                    readBuffer.checkpoint(new CheckpointState(records.size()));
                 }
```
wrijeff, line 59: What if this records list is empty, say in the event of Peer Forwarder forwarding all of the records?

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
chenqi0805: Sure.

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
chenqi0805: This part largely refers to `LinkedListBlockingQueue.offer(E e, long timeout, TimeUnit unit)` in which there is only lock release in the finally block. TBH, I cannot tell the essential difference between placing it here and placing it in finally. Is there any good practice I missed here?

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
wrijeff: Correct, we'll have to make it clear that a buffer's capacity is only freed up once an item has been both _read and 
 acknowledged_. (Present behavior is only after an item is _read_). 

We can revisit later if we think having granular control over both internal queue capacity and in-flight capacity is necessary, but for now combining the capacities keeps things simple.

wrijeff (changes requested):
My original understanding was that the buffer would provide a checkpoint object when a consumer calls `read()`. Once the consumer is done processing all records from the batch, it would use the checkpoint object somehow to call back to the buffer acknowledging that the records have been processed and are no longer "in-flight". Essentially the checkpoint object is a read receipt. Simply reading an item from the buffer would no longer free up space for more writes; the new buffer logic would only free up space when an item has been _read and acknowledged_ by the reader.

A good starting point would be to use a [semaphore](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html). It handles most of the heavy synchronization and signaling for us. Another benefit we get from the semaphore is thread fairness, enabled by a boolean on the constructor (might have a performance impact)
1. On buffer creation we would create a semaphore with capacity equal to the buffer size
2. On `write()`, `tryAcquire(long timeout, TimeUnit unit)`. This will handle the timer/timeout logic for us
3. On `read()`, return a checkpoint object with the number of items read (the receipt)
4. Onnce finished processing, the consumer is responsible for interacting with the buffer/checkpoint object which ultimately calls `release(int permits)` on the semaphore

In this current PR, I see ProcessWorker creating a CheckpointState from scratch to pass back to the buffer. We don't want consumers to have to know how to create buffer-specific objects, so having `read()` call return the checkpoint object (potentially as a pair <items, checkpoint> or via some other signature) could be an improvement.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
chenqi0805: Great point! I think in this case we do need to apply your idea on passing the Checkpoint obj to consumer.

AustinTag: Oh actually, this is a blocker. There is no notion that the number of output records is equal to the number of input records. We should wait for this to complete and checkpoint with the number of records read from the buffer. We don't care how many are output at the end.

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
AustinTag: So there is no functional difference here because it seems that there isn't a way for an error to be thrown. The practice is that if there is an action you want to make sure happens (such as unlocking a lock), then we want that to be done in the finally block. 

In addition, I'm a little unsure on why we need the write function to unblock threads. Is the checkpoint calling the notFull signal not sufficient here?

chenqi0805: > In addition, I'm a little unsure on why we need the write function to unblock threads. Is the checkpoint calling the notFull signal not sufficient here?

So currently the checkpoint API calls notFull.signal ONLY when the previous capacity is full. In theory, there could be more than one thread waiting for write into the buffer due to previous full capacity (although in practice looks like our source that executes write is always in the main thread). That is why I called notFull.signal() here to wake up another waiting thread if the capacity is not full yet. Otherwise those threads will never be waken up.

I guess alternatively, we could just invoke notFull.signal unconditionally in the checkpoint API. I am not sure which one is better in performance.

> Also, why not call the "signalNotFull()" function here if we need to call this signal?

so here we already hold the guarding lock while signalNotFull includes grabbing the lock.

AustinTag: I think calling the signal in the checkpoint whether or not it was previously full makes more sense to me. The reason is that the write can only increase the buffer capacity. So it doesn't make sense to send the notFull signal there, because the buffer can never change from full to not full there.

chenqi0805: Thanks for the input here. I am trying to simplify the logic here by using semaphore as Jeff suggested. So this part might be rewritten.

chenqi0805:
> In this current PR, I see ProcessWorker creating a CheckpointState from scratch to pass back to the buffer. We don't want consumers to have to know how to create buffer-specific objects, so having read() call return the checkpoint object (potentially as a pair <items, checkpoint> or via some other signature) could be an improvement.

@wrijeff Upon a second thought, I would like to push back the improvement of let read() return the checkpoint object in a separate PR for the following reason:

(1) For now CheckpointState is in data-prepper-api which is supposed to be visible to all plugins. So letting ProcessWorker construct it from scratch is still valid as we expect it is the data-prepper level knowledge. We can doublecheck later if the model should be restricted to buffer plugin only.

(2) This enhancement on read() will touch quite a lot of buffer code not directly relevant to checkpoint. So I would prefer doing it incrementally.

chenqi0805:
@AustinTag @wrijeff I switched the implementation of doWrite and checkpoint based off semaphore, which avoids low level synchronization operations and should have made the logic much simpler. Now it is ready for another look except for the comment above ^^. Thanks for all the input.

data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/buffer/BlockingBufferTests.java
```diff
@@ -105,6 +138,34 @@ public void testBatchRead() throws Exception {
         }
     }
 
+    @Test
+    public void testWriteAfterCheckedBatchReads() throws Exception {
```
wrijeff, line 142: How is this different than the `testWriteIntoEmptySpaceAfterCheckedRead` above? Aren't both filling a buffer, reading, checkpointing, then asserting writes succeed?

wrijeff (commented):
Posted a few minor comments/questions.

Regarding letting the ProcessWorker create a Checkpoint object instead of using one returned by the buffer:

That's fine for now, we can revisit as you said. Both approaches (manually creating a Checkpoint or using the one returned by the buffer) still introduce a temporal coupling, in that callers of the buffer are _expected_ to call `read()` then `checkpoint()` otherwise the buffer gets into a bad state. 

We might consider something like changing the `read()` signature to take in a checkpoint object in the future to enforce a stronger coupling.

data-prepper-plugins/common/src/test/java/com/amazon/dataprepper/plugins/buffer/BlockingBufferTests.java
chenqi0805: Will remove this one. This test case was deviced to ensure full coverage in the previous implementation, even so it was over-stretching.

wrijeff (approved):

chenqi0805:
> That's fine for now, we can revisit as you said. Both approaches (manually creating a Checkpoint or using the one returned by the buffer) still introduce a temporal coupling, in that callers of the buffer are expected to call read() then checkpoint() otherwise the buffer gets into a bad state.

@wrijeff Upon temporal coupling issue between read and checkpoint, it makes me think of internalize checkpointState as an offset-like variable in the Buffer Plugin and remove checkpointState from the signature of checkpoint API. That way we will never mess up the internal state of BlockingBuffer by (1) invoking read and checkpoint in the wrong order (2) invoking checkpoint with an improperly constructed checkpointState

AustinTag (approved):

PR #291, opened 2021-01-28
ENH: decouple read and checkpoint APIs temporally in BlockingBuffer

chenqi0805:
*Issue #, if available:*
#216 

*Description of changes:*
This PR

(1) Removes explicit CheckpointState and refactor the checkpoint state variables into BlockingBuffer (more generically the buffer plugin is responsible for its checkpoint state variables)
(2) Simplifies the checkpoint API signature.
(3) Modified doRead in BlockingBuffer to keep track of the offset to be checked while polling from the blockingQueue.

Note: with this change, readBuffer.read() and readBuffer.checkpoint() are independent in ordering, i.e. messing up the temporal call order should not cause BlockingBuffer enter into an invalid state.


data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -112,12 +113,17 @@ public void doWrite(T record, int timeoutInMillis) throws TimeoutException {
         final Stopwatch stopwatch = Stopwatch.createStarted();
         try {
             while (stopwatch.elapsed(TimeUnit.MILLISECONDS) < timeoutInMillis && records.size() < batchSize) {
-                final T record = blockingQueue.poll(timeoutInMillis, TimeUnit.MILLISECONDS);
-                if (record != null) { //record can be null, avoiding adding nulls
-                    records.add(record);
-                }
-                if (records.size() < batchSize) {
-                    blockingQueue.drainTo(records, batchSize - records.size());
+                synchronized (this) {
```
kowshikn, line 116: Why should this call be synchronized? Both the blocking queue poll and recordsToBeCheckpointed both are thread safe.

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -112,12 +113,17 @@ public void doWrite(T record, int timeoutInMillis) throws TimeoutException {
...
+                    final T record = blockingQueue.poll(timeoutInMillis, TimeUnit.MILLISECONDS);
+                    if (record != null) { //record can be null, avoiding adding nulls
+                        records.add(record);
+                    }
+                    numRecordsToBeChecked.incrementAndGet();
```
kowshikn, line 121: This should be inside the `if(records!=null)`. Correct?

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -112,12 +113,17 @@ public void doWrite(T record, int timeoutInMillis) throws TimeoutException {
...
+                    if (records.size() < batchSize) {
+                        final int maxRemaining = batchSize - records.size();
+                        blockingQueue.drainTo(records, maxRemaining);
+                        numRecordsToBeChecked.addAndGet(maxRemaining);
```
kowshikn, line 125: This should be `numRecordsToBeChecked.addAndGet(recordsSizeBeforeDrain-recordsSizeAfterDrain)`. Correct?

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
yadavcbala: +1

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
```diff
@@ -140,8 +146,8 @@ public static PluginSetting getDefaultPluginSettings() {
     }
 
     @Override
-    public void checkpoint(final CheckpointState checkpointState) {
-        final int numCheckedRecords = checkpointState.getNumCheckedRecords();
+    public void checkpoint() {
+        final int numCheckedRecords = numRecordsToBeChecked.getAndSet(0);
```
kowshikn, line 150: I am not sure how this will work. numRecordsToBeChecked is an attribute of the BlockingBuffer, without the state you can not decrement or set to zero.

An example,

Lets assume the buffer size is 100, batch size 50 and we have two processors (aka processworker threads)

1. processor-1  reads 50 records, numRecordsToBeChecked is set to 50.
2. processor-2  reads 50 records,  numRecordsToBeChecked is set to 50.
3. processor-1 checkpoints, numRecordsToBecheckec is set to 0.

So, processor-2 is still working, but the buffer is read to accept more records.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
```diff
@@ -58,7 +56,7 @@ public void run() {
                     postToSink(records); //TODO use the response to ack the buffer on failure?
```
yadavcbala, line 56: u may also want to remove this TODO

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
chenqi0805: Thanks for the catch

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/ProcessWorker.java
```diff
@@ -58,7 +56,7 @@ public void run() {
...
                 }
                 // Checkpoint the current batch read from the buffer after being processed by prepper and sinks.
-                readBuffer.checkpoint(checkpointState);
+                readBuffer.checkpoint();
```
yadavcbala, line 59: this works because we currently do not worry about partially processed records - per design we expect some preppers to drop some records [based on context - it may mean they are processed] and we indefinitely wait for sinks to post and fail even if one of the sink fails. I do not see we changing any of this, so we are good here

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/buffer/BlockingBuffer.java
chenqi0805: Thanks for the crucial catch. Now this PR breaks the previously desired functionally with checkpoint API.

chenqi0805:
@yadavcbala @kowshikn Thanks for the review. Discussed with Kowshik offline, this PR is not doing what we want in terms of checkpoint API, so I will close this one. Instead, will submit a new PR that let buffer.read returns the checkpointState object: 

```
Map.Entry<Collection,CheckpointState> bufferResponse = readBuffer.read(pipeline.getReadBatchTimeoutInMillis());
```

So that buffer will create its own checkpointState class based on its implementation.

PR #294, opened 2021-01-30
ENH: metrics in AbstractBuffer

chenqi0805:
*Issue #, if available:*

*Description of changes:*
This PR introduces the following metrics in AbstractBuffer

- recordsInflightCounter
- recordsProcessedCounter (checkpointed records)
- checkpointTimer

and removes the recordsReadCounter as now it breaks down into recordsInflightCount + recordsProcessedCount

### DIFF
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
@@ -46,6 +46,11 @@ public class MetricNames {
     public static final String READ_TIME_ELAPSED = "readTimeElapsed";
 
     /**
+     * Metric representing the time elapsed while checkpointing
+     */
+    public static final String CHECKPOINT_TIME_ELAPSED = "checkpointTimeElapsed";
+
+    /**
      * Metric representing the count of write timeouts to a Buffer
      */
     public static final String WRITE_TIMEOUTS = "writeTimeouts";
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
@@ -26,7 +26,7 @@ public class MetricNames {
     public static final String RECORDS_WRITTEN = "recordsWritten";
 
     /**
-     * Metric representing the number of records read from a buffer
+     * Metric representing the number of records read from a buffer.
      */
     public static final String RECORDS_READ = "recordsRead";
 
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
@@ -26,6 +26,11 @@ public class MetricNames {
     public static final String RECORDS_WRITTEN = "recordsWritten";
 
     /**
+     * Metric representing the number of records read from a buffer.
+     */
+    public static final String RECORDS_READ = "recordsRead";
+
+    /**
      * Metric representing the number of records read from a buffer but unchecked.
      */
     public static final String RECORDS_INFLIGHT = "recordsInflight";
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
@@ -33,12 +33,12 @@ public class MetricNames {
     /**
      * Metric representing the number of records read from a buffer but unchecked.
      */
-    public static final String RECORDS_INFLIGHT = "recordsInflight";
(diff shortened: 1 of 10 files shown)


data-prepper-api/src/main/java/com/amazon/dataprepper/model/buffer/AbstractBuffer.java
```diff
@@ -76,10 +80,18 @@ public void write(T record, int timeoutInMillis) throws TimeoutException {
...
-        recordsReadCounter.increment(readResult.getKey().size()*1.0);
+        recordsInflightCounter.increment(readResult.getValue().getNumRecordsToBeChecked()*1.0);
         return readResult;
     }
 
+    @Override
+    public void checkpoint(final CheckpointState checkpointState) {
+        checkpointTimer.record(() -> doCheckpoint(checkpointState));
+        final int numRecordsToBeChecked = checkpointState.getNumRecordsToBeChecked();
+        recordsInflightCounter.increment(-numRecordsToBeChecked);
```
AustinTag, line 91: counters can only go up: https://micrometer.io/docs/concepts#_counters. while this may work in a unit test because we are using a SimpleRegistry which keeps the latest value in memory only, this will not work in production.

data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
```diff
@@ -26,9 +26,14 @@ private MetricNames() {}
     public static final String RECORDS_WRITTEN = "recordsWritten";
 
     /**
-     * Metric representing the number of records read from a buffer
+     * Metric representing the number of records read from a buffer but unchecked.
      */
-    public static final String RECORDS_READ = "recordsRead";
```
AustinTag, line 31: Why are we removing this metric? Records written and records read seem like useful metrics.

data-prepper-api/src/main/java/com/amazon/dataprepper/model/buffer/AbstractBuffer.java
AustinTag: this will need to be a gauge

data-prepper-api/src/main/java/com/amazon/dataprepper/model/buffer/AbstractBuffer.java
```diff
@@ -17,10 +17,12 @@
 public abstract class AbstractBuffer<T extends Record<?>> implements Buffer<T> {
     protected final PluginMetrics pluginMetrics;
     private final Counter recordsWrittenCounter;
-    private final Counter recordsReadCounter;
+    private final Counter recordsInflightCounter;
+    private final Counter recordsProcessedCounter;
```
AustinTag, line 21: I think the name of this metric is correct, it is a count of the records processed by this pipeline. However, it would be nice if this could be a pipeline level metric rather than a buffer level metric. 

pipelineName.buffer.recordsProcessed seems to imply that it is a count of the records processed by that buffer, rather than the records processed by the pipeline.

recordsRead, recordsWritten, read/write latency, and inflight are all specific to the buffer. Records processed refers to records processed by the pipeline

AustinTag (changes requested):

data-prepper-api/src/main/java/com/amazon/dataprepper/model/buffer/AbstractBuffer.java
chenqi0805: In our case, although inflight could be specific to buffer, it still needs to be updated when recordsProcessed is updated. So we need either (1) expose recordsInflightCounter to the popeline or (2) make recordsInflightCounter a pipeline level metric.

chenqi0805: Or what if we still keep it as a buffer level metric but change the name into pipelineName.buffer.recordsChecked?

data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
chenqi0805: Added back-in

data-prepper-api/src/main/java/com/amazon/dataprepper/model/buffer/AbstractBuffer.java
kowshikn: Do we have a conclusion on this discussion? Asking because I missed the offline conversation..

chenqi0805: Yes. The solution is to make the prefix of the metric name reflect the pipeline level, i.e. `pipelineName.recordsProcessed` instead of `pipelineName.bounded_buffer.recordsProcessed`

kowshikn (approved):

PR #297, opened 2021-02-02
Added initial EC2 CFN template.

wrijeff:
*Description of changes:*

Adding initial CFN template. Tested against a public AWS ES instance with an IAM role access policy. Still need to test w/ VPC, and potentially update the Data Prepper config default values.

### DIFF
--- a/deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
+++ b/deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
@@ -5,30 +5,30 @@ Parameters:
   AWSElasticsearchEndpoint:
     Description: Endpoint of the AWS Elasticsearch Service domain (including https://)
     Type: String
-    AllowedPattern: .+
-    ConstraintDescription: cannot be empty
+    AllowedPattern: https:\/\/[a-z0-9-\.]+(es.amazonaws.com)
+    ConstraintDescription: must be a valid Amazon Elasticsearch Service domain endpoint starting with https://
   AWSElasticsearchRegion:
     Description: Region of the AWS Elasticsearch Service domain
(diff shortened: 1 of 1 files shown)


laneholloway (commented):
This looks good, do we expect to provide more stringent constraints on many of the parameters?

wrijeff:
> This looks good, do we expect to provide more stringent constraints on many of the parameters?

Ah good point, I glossed over the parameters early on just to get the template working. Can easily add some better validation while this PR is still out for review.

kowshikn (approved):
> > This looks good, do we expect to provide more stringent constraints on many of the parameters?
> 
> Ah good point, I glossed over the parameters early on just to get the template working. Can easily add some better validation while this PR is still out for review.

Not sure what validations we can do with existing parameters. Otherwise this LGTM for the first version. `cfn-hup` should be included in the next version or release, as thats more useful.

laneholloway (approved):
For the first version this is fine, but there are simple validations we can do to ensure a true http/s address, etc. in the future.

wrijeff:
> > > This looks good, do we expect to provide more stringent constraints on many of the parameters?
> > 
> > 
> > Ah good point, I glossed over the parameters early on just to get the template working. Can easily add some better validation while this PR is still out for review.
> 
> Not sure what validations we can do with existing parameters. Otherwise this LGTM for the first version. `cfn-hup` should be included in the next version or release, as thats more useful.

+1 cfn-hup, saw that in other templates but deprioritized it for this initial version.

Re: validation - I'll be publishing a commit w/ regexes to ensure the fields match up with their types (e.g. ES endpoint starts with https://, IAM roles only contain allowed characters, etc.)

PR #298, opened 2021-02-03
Added Jaeger HotROD on EC2 setup script.

wrijeff:
*Description of changes:*
We want to allow users to quickly setup a sample Jaeger app to emit trace data for Data Prepper to consume so they can try out the new trace analytics plugin.

* Added an /examples/aws directory for AWS-specific content
* Added a single bash script to install necessary components for running the HotROD app via docker-compose
  * Users will run `sh setup-jaeger-hotrod.sh` to stand up a working application
* I'll add readme's to the new aws and jaeger directories in a follow-up. 

The alternative to a setup script is giving the user a series of steps like:
1. Install Docker
1. Install Docker Compose
1. Clone this repo
1. Navigate to /examples/aws/<something>
1. Run `docker-compose up -d`

This setup script is extremely focused and hides the technical details, but I think that's ok for this usecase. Open to thoughts.

### DIFF
--- a/examples/aws/jaeger-hotrod-on-ec2/setup-jaeger-hotrod.sh
+++ b/examples/aws/jaeger-hotrod-on-ec2/setup-jaeger-hotrod.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+# Script to install and run the Jaeger Hot R.O.D. sample application.
+# Installs docker, docker-compose, writes configuration files, then runs 'docker-compose up'
+
+sudo yum install docker -y
+sudo usermod -aG docker $USER
+sudo systemctl start docker
+sudo curl -L "https://github.com/docker/compose/releases/download/1.28.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+sudo chmod +x /usr/local/bin/docker-compose
+
+cat <<EOT >> otel-collector-config.yml
+receivers:
+  jaeger:
+    protocols:
+      grpc:
+
+exporters:
+  otlp/2:
+    endpoint: localhost:21890
+    insecure: true
+  logging:
+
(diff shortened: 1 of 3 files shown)


examples/aws/jaeger-hotrod-on-ec2/setup-jaeger-hotrod.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+# Script to install and run the Jaeger Hot R.O.D. sample application.
+# Installs docker, docker-compose, writes configuration files, then runs 'docker-compose up'
```
laneholloway, line 3: Do we want to provide details as to where it installs all of the components? (not that they couldn't read the code script and find it...)

laneholloway (approved):
I'd like to over-index on explanation within our scripts so people know exactly what it is doing by reading the comments within it. But, that is a larger discussion that we can have and it can be updated later.

wrijeff:
> I'd like to over-index on explanation within our scripts so people know exactly what it is doing by reading the comments within it. But, that is a larger discussion that we can have and it can be updated later.

Good point, probably best to provide more docs than less in externally facing projects. Will revise.

examples/aws/jaeger-hotrod-on-ec2/setup-jaeger-hotrod.sh
wrijeff: Added in-line comments to make the script a bit more readable. Also added a README to the directory which links to the various components used here.

laneholloway (approved):
Thanks for the comments!

wrijeff:
@kowshikn: Merging this now as is, we can cut down some of the scope later today if we want to

PR #300, opened 2021-02-04
Added support for ES domains in VPCs to the CloudFormation script.

wrijeff:
*Description of changes:*
Added support for ES domains in VPCs to the CloudFormation script. Tested against public and VPC ES domains.

### DIFF
--- a/deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
+++ b/deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
@@ -13,6 +13,9 @@ Parameters:
     AllowedPattern: "[a-z]+-[a-z]+-[0-9]+"
     Default: us-east-1
     ConstraintDescription: must be a valid AWS region (e.g. us-west-2)
+  AWSElasticsearchSubnetId:
+    Description: The subnet ID of the AWS Elasticsearch Service domain (Leave blank if the domain is not in a VPC)
+    Type: String
   DataPrepperVersion:
     Description: Version of Data Prepper to download and run
     Type: String
@@ -47,6 +50,9 @@ Parameters:
     Default: 0.0.0.0/0
     AllowedPattern: (\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})/(\d{1,2})
     ConstraintDescription: must be a valid IP CIDR range of the form x.x.x.x/x.
+Conditions:
+  DomainIsPublic: !Equals [!Ref AWSElasticsearchSubnetId, ""]
+  DomainIsInVPC: !Not [Condition: DomainIsPublic]
 Resources:
   EC2Instance:
     Type: AWS::EC2::Instance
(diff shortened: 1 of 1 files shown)


deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
```diff
@@ -13,6 +13,9 @@ Parameters:
     AllowedPattern: "[a-z]+-[a-z]+-[0-9]+"
     Default: us-east-1
     ConstraintDescription: must be a valid AWS region (e.g. us-west-2)
+  AWSElasticsearchSubnetId:
+    Description: The subnet ID of the AWS Elasticsearch Service domain (Leave blank if the domain is not in a VPC)
+    Type: String
```
laneholloway, line 18: Are there any constraints on this?

laneholloway (approved):
If there are constraints for the subnet id, lets add them in.

deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
```diff
@@ -13,6 +13,9 @@ Parameters:
     AllowedPattern: "[a-z]+-[a-z]+-[0-9]+"
     Default: us-east-1
     ConstraintDescription: must be a valid AWS region (e.g. us-west-2)
+  AWSElasticsearchSubnetId:
+    Description: The subnet ID of the AWS Elasticsearch Service domain (Leave blank if the domain is not in a VPC)
```
chenqi0805, line 17: Could also add an e.g. as above

deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
wrijeff: From what I can tell no - the AWS console only indicates a limit of 256 characters. To test it out I made a subnet with weird characters named ` ######$$$$$$!!!!!!@@-----___++` which was created successfully. For that reason I didn't think it was worth adding a constraint.

wrijeff: It is worth noting that CFN offers a Subnet parameter which populates a drop-down box of existing subnet IDs to choose from, however it can't be left blank. 

So if we were to create separate VPC/no-VPC templates we could use the Subnet parameter, but not in the combined template as it currently stands.

laneholloway: yeah, not worth it. I'm good with it.

deployment-template/ec2/data-prepper-ec2-deployment-cfn.yaml
wrijeff: Unfortunately subnets can be named whatever (the only constraint is a 256 character limit), so we can't really offer a meaningful example

chenqi0805: I was thinking some generic pattern like `subnet-4b18e22d(172.31.0.0/20)|us-east-1b` from the dropdown menu. But not sure if it is the subnet ID. Just to give cx a manifestation.

chenqi0805 (approved):

PR #307, opened 2021-02-08
Reformatting OTel Trace Raw Prepper and Blocking Buffer README

sshivanii:
*Issue #, if available:*

*Description of changes:*
- This PR reformats otel_trace_raw_prepper README.
- Moves the Blocking Buffer out of the common subproject into a separate subproject under data-prepper-plugins/blocking-buffer. The new blocking-buffer subproject has it's own reformatted README and gradle build file.
- Updates all usages of Blocking Buffer in the codebase with the new path.


settings.gradle
```diff
@@ -24,4 +24,6 @@ include 'data-prepper-benchmarks:service-map-stateful-benchmarks'
 include 'data-prepper-plugins:otel-trace-raw-prepper'
 include 'data-prepper-plugins:otel-trace-source'
 include 'data-prepper-plugins:peer-forwarder'
+include 'data-prepper-plugins:blocking-buffer'
+findProject(':data-prepper-plugins:blocking-buffer')?.name = 'blocking-buffer'
```
chenqi0805, line 28: Is this needed?

data-prepper-plugins/blocking-buffer/README.md
```diff
@@ -0,0 +1,20 @@
+# Blocking Buffer
+
+This is a buffer based off `LinkedBlockingQueue` bounded to the specified capacity. One can read and write records with specified timeout value.
+
+## Configuration
+- buffer_size => An `int` representing max number of records the buffer accepts.
```
chenqi0805, line 6: Please also specify the default value. Same below for `batch_size`.

chenqi0805: Due to the addition of checkpoint API, here we need to specify that it represents the maximum number of unchecked records the buffer accepts, where
```
# of unchecked records = # of records written into the buffer + # of in-flight records (i.e. records read from the buffer but not yet checked by invoking checkpoint API).
```

data-prepper-plugins/blocking-buffer/README.md
```diff
@@ -0,0 +1,20 @@
...
+- batch_size => An `int` representing max number of records the buffer returns on read.
+
+## Usages
```
chenqi0805, line 9: Place usages section before configuration.

data-prepper-plugins/otel-trace-raw-prepper/README.md
```diff
@@ -2,7 +2,22 @@
...
 This is a prepper that serializes collection of `ExportTraceServiceRequest` sent from [otel-trace-source](../dataPrepper-plugins/otel-trace-source) into collection of string records. 
 
+## Usages
+Example `.yaml` configuration
 ```
 prepper:
     - otel_trace_raw_prepper:
-```
\ No newline at end of file
+```
```
chenqi0805, line 10: We can still add `## Configuration` title as a placeholder.

data-prepper-plugins/blocking-buffer/README.md
```diff
@@ -0,0 +1,20 @@
...
+Example `.yaml` configuration
+```
+buffer:
+    - bounded_blocking:
+```
+*Note*: *By default, Data Prepper uses only one buffer. the `bounded_blocking` buffer, so this section in the `.yaml` need not be defined unless one wants to mention a custom buffer or tune the buffer settings* 
+
```
chenqi0805, line 16: Although there is no custom metrics, we still need to add the Metrics section to point out the buffer inherits common metrics in `AbstracBuffer`.

data-prepper-plugins/common/build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
 }
 dependencies {
     compile project(':data-prepper-api')
+    compile project(':data-prepper-plugins:blocking-buffer')
```
yadavcbala, line 6: can you confirm why it needs to be here - Are we using this as part of any other common plugin ?

data-prepper-plugins/blocking-buffer/README.md
sshivanii: Will add this

settings.gradle
sshivanii: It was automatically added when the blocking-buffer subproject was created, I'll check if the findProject is needed or not.

data-prepper-plugins/common/build.gradle
sshivanii: Yes, the BufferFactoryTests and FileSourceTests are using the BlockingBuffer.

yadavcbala: I think if its just for tests - we should consider moving tests related BlockingBuffer to its specific project. And for other tests, we should consider using a DummyBuffer over this. So, that way if we update Plugin architecture to Service Loader interface, we can decouple these completely. Not a must but since we are moving them lets move them completely.

chenqi0805: @yadavcbala Thanks for the catch! Did not notice that. Yes, I agree that we should use DummyBuffer for BufferFactoryTests and FileSourceTests. There is already `TestBuffer` in the buffer test utlilities which should be applicable to FileSourceTests. Not sure if we can annotate that one for `BufferFactoryTests`.

sshivanii: I agree, went ahead with using `TestBuffer` in `FileSourceTests` and discussed offline with @yadavcbala to proceed with annotating `TestBuffer` for usage in `BufferFactoryTests`.

chenqi0805 (approved):
LGTM!

yadavcbala (approved):

Issue #309, opened 2021-02-09
Track dependency CVEs and version updates


wrijeff:
Need a way to ensure our dependencies stay up-to-date, especially to patch reported CVEs in a timely fashion.

wrijeff:
Automated vulnerability reports: https://docs.github.com/en/github/managing-security-vulnerabilities/about-alerts-for-vulnerable-dependencies

Dependency upgrades: https://docs.github.com/en/github/administering-a-repository/about-dependabot-version-updates

wrijeff:
You can see example automated PRs in my fork: https://github.com/wrijeff/simple-ingest-transformation-utility-pipeline/pulls

PR #310, opened 2021-02-09
Adding Dependabot for automated dependency upgrades.

wrijeff:
*Issue #, if available:* #309

*Description of changes:*
* Adding dependabot config file to enable automated dependency upgrades 
  * Docs: https://docs.github.com/en/github/administering-a-repository/about-dependabot-version-updates
  * How to configure: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/metrics/MetricNames.java
@@ -26,9 +26,19 @@ public class MetricNames {
     public static final String RECORDS_WRITTEN = "recordsWritten";
 
     /**
-     * Metric representing the number of records read from a buffer
+     * Metric representing the number of records read from a buffer.
      */
     public static final String RECORDS_READ = "recordsRead";
+
+    /**
+     * Metric representing the number of records read from a buffer but unchecked.
+     */
+    public static final String RECORDS_INFLIGHT = "recordsInFlight";
+
+    /**
+     * Metric representing the number of records read from a buffer and processed by the pipeline.
+     */
+    public static final String RECORDS_PROCESSED = "recordsProcessed";
 
     /**
      * Metric representing the time elapsed while writing to a Buffer
@@ -39,6 +49,11 @@ public class MetricNames {
      * Metric representing the time elapsed while reading from a buffer
      */
     public static final String READ_TIME_ELAPSED = "readTimeElapsed";
+
+    /**
+     * Metric representing the time elapsed while checkpointing
+     */
+    public static final String CHECKPOINT_TIME_ELAPSED = "checkpointTimeElapsed";
 
     /**
      * Metric representing the count of write timeouts to a Buffer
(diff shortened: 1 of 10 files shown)


wrijeff:
You can see example automated PRs in my fork: https://github.com/wrijeff/simple-ingest-transformation-utility-pipeline/pulls

laneholloway (approved):
This looks handy.

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gradle
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 3
```
yadavcbala, line 7: does it update for all branches ? if so, should we avoid for released branches ? [just asking ]

yadavcbala:
Thanks for doing this, this definitely is handy with Insights and other associated features it brings in [in addition to dependency versions]

.github/dependabot.yml
wrijeff: Good question. Currently it checks only the default branch, but [it can be configurable](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#target-branch) if we start using a development branch process in the future.

yadavcbala (approved):

kowshikn:
1. We can not do major/minor(sometimes) version changes always, I know we have the integration tests that builds and tests data prepper to an extent, so it should catch them but I am worried about OpenTelemetry because they could mess the protobuf between versions. Definitely patch is a good one to do.
2.  Our projiect uses `versionMap` does it support that? 
3. We should skip the gradle java app inside the `examples`, we have no integ test to catch it.

wrijeff:
> 1. We can not do major/minor(sometimes) version changes always, I know we have the integration tests that builds and tests data prepper to an extent, so it should catch them but I am worried about OpenTelemetry because they could mess the protobuf between versions. Definitely patch is a good one to do.
> 2. Our projiect uses `versionMap` does it support that?
> 3. We should skip the gradle java app inside the `examples`, we have no integ test to catch it.

@kowshikn: great points.
1. Yep, for some of our less-stable dependencies we should probably `ignore` those for now. I can add otel proto to the config file and we can watch for automated upgrades that look questionable. And yes we will need to continue to invest in our automated testing to catch issues caused by version upgrades.
2. Version map - I don't think so. But if all dependencies are automatically bumped to the latest then they should remain consistent across all packages (as all will be bumped to the same version).
3. Good catch. Unfortunately it doesn't look like we can ignore just one directory, so we'll instead have to list out all the directories we _do_ want scanned.

I feel this tool does more good than harm overall (with a bit of effort on our part to integrate it) - I can move this to a new "dependencies" branch where we can slowly bring in each package as we get more comfortable with the tool. Thoughts?

kowshikn:
I absolutely agree that this tool is more good. I have no concerns adding it even to the master branch because there is always a PR and I think we can find issues and the our integ test is very solid at least for **Trace Analytics**. 
 
The real issue is the how we build our dependency in this project, its not something I am very happy about. I am sure we can fix some of them soon when we address the reduce build/docker image size.

PR #311, opened 2021-02-09
Fixed minor typo in log4j properties file.

wrijeff:
*Description of changes:*
1. Fixed typo in log4j properties file (`plugin` to `plugins`)
2. Reduced logging threshold from ERROR to WARN to not hide Armeria warnings
    * I didn't see an increase in log noise, but if it does become too noisy then we can just elevate `com.linecorp.armeria.client` instead
3. Fixed the dev example app not passing in the log4j file on startup

### DIFF
--- a/shared-config/log4j.properties
+++ b/shared-config/log4j.properties
@@ -1,5 +1,5 @@
 # Logging Level
-log4j.rootLogger=ERROR, CONSOLE
+log4j.rootLogger=WARN, CONSOLE
 log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
 log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
 log4j.appender.CONSOLE.Threshold=INFO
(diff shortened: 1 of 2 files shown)


shared-config/log4j.properties
```diff
@@ -17,4 +17,4 @@ log4j.appender.file.layout.ConversionPattern=%d{ISO8601} [%t] %-5p %40C - %m%n
 #package specific levels
 log4j.logger.com.amazon.dataprepper.pipeline=INFO
 log4j.logger.com.amazon.dataprepper.parser=INFO
-log4j.logger.com.amazon.dataprepper.plugin=INFO
+log4j.logger.com.amazon.dataprepper.plugins=INFO
```
yadavcbala, line 20: I have kept fixing this in the images i uploaded to ECR and always postponed to push a PR. Thanks for doing this

wrijeff: I have also been meaning to fix this for awhile but have been pushing it off too 😉

chenqi0805 (approved):

yadavcbala (approved):

Issue #221, opened 2020-12-11
Flaky testcase in OtelTraceSource


chenqi0805:
Occasionally found in CI:
```
> Task :data-prepper-plugins:otel-trace-source:test

OTelTraceSourceTest > testBufferFull() FAILED
    io.grpc.StatusRuntimeException at OTelTraceSourceTest.java:71


7 tests completed, 1 failed
> Task :data-prepper-plugins:otel-trace-source:test FAILED
```
@AustinTag has also seen this in CI on PR.

kowshikn:
Is this after the latest fix from @AustinTag ?

sshivanii:
Encountered the same on my PR.

chenqi0805:
It might be due to UnprocessedRequestException or Request timeout exception in the armeria client. As a workaround, I will try adding retry to see if it stabilizes.

chenqi0805:
adding retry does not prevent it. Need to investigate further.

chenqi0805:
Was just able to get a full Traceback:

```
2020-12-17T21:34:15.6198776Z 2020-12-17T21:34:15.593+0000 [DEBUG] [TestEventLogger] OTelTraceSourceTest > testBufferFull() FAILED
2020-12-17T21:34:15.6200032Z 2020-12-17T21:34:15.594+0000 [DEBUG] [TestEventLogger]     io.grpc.StatusRuntimeException: CANCELLED
2020-12-17T21:34:15.6201785Z 2020-12-17T21:34:15.594+0000 [DEBUG] [TestEventLogger]         at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262)
2020-12-17T21:34:15.6203663Z 2020-12-17T21:34:15.594+0000 [DEBUG] [TestEventLogger]         at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243)
2020-12-17T21:34:15.6205770Z 2020-12-17T21:34:15.594+0000 [DEBUG] [TestEventLogger]         at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156)
2020-12-17T21:34:15.6208412Z 2020-12-17T21:34:15.594+0000 [DEBUG] [TestEventLogger]         at io.opentelemetry.proto.collector.trace.v1.TraceServiceGrpc$TraceServiceBlockingStub.export(TraceServiceGrpc.java:201)
[...]

PR #323, opened 2021-02-10
221: Removing unnecessary static keyword in OTelTraceSource test class.

wrijeff:
*Issue #, if available:* #221

*Description of changes:*
* Changing `BlockingBuffer` variable in unit test from static to instance

Potentially fixing the flaky testBufferFull() test (though I'm not 100% sure as it's tough to reproduce). Noticed the buffer object isn't being re-initialized between each test, so maybe the buffer is already full by the time this particular test is reached?

### DIFF
--- a/data-prepper-plugins/otel-trace-source/src/test/java/com/amazon/dataprepper/plugins/source/oteltrace/OTelTraceSourceTest.java
+++ b/data-prepper-plugins/otel-trace-source/src/test/java/com/amazon/dataprepper/plugins/source/oteltrace/OTelTraceSourceTest.java
@@ -63,6 +63,8 @@ public class OTelTraceSourceTest {
     PluginSetting pluginSetting;
     PluginSetting testPluginSetting;
 
+    private BlockingBuffer<Record<ExportTraceServiceRequest>> buffer;
+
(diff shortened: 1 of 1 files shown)


data-prepper-plugins/otel-trace-source/src/test/java/com/amazon/dataprepper/plugins/source/oteltrace/OTelTraceSourceTest.java
```diff
@@ -73,7 +75,6 @@
                     .addInstrumentationLibrarySpans(InstrumentationLibrarySpans.newBuilder()
                             .addSpans(Span.newBuilder().setTraceState("FAILURE").build())).build()).build();
     private static TraceServiceGrpc.TraceServiceBlockingStub CLIENT;
-    private static final BlockingBuffer<Record<ExportTraceServiceRequest>> BUFFER = getBuffer();
```
kowshikn, line 76: This change makes sense, but I doubt if this the cause. :D

kowshikn (approved):

sshivanii (approved):
Change makes sense. LGTM!

Issue #321, opened 2021-02-10
Enable TLS over /metrics endpoint


wrijeff:
Following AppSec review, TLS is sort of mandatory for all APIs and avoiding it requires approval. Considering we have time till March 1st week, we will implement the TLS for metrics server.

-  Use the [data-prepper configuration](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/master/docs/readme/project_setup.md#data-prepper-configuration) to specify the ssl flag and the certs. The default should be true. 

- Make sure the configuration is pushed to the bootstrap code
https://github.com/opendistro-for-elasticsearch/data-prepper/blob/c67c6dd7454b4071f82e7e080b46901141d1145c/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java#L16-L25

- Update Monitoring read me with ssl setup instruction. (Our monitoring readme lacks lot of detail. We should have more clear instructions of how to setup grafana and show some pictures.)

PR #327, opened 2021-02-12
321: Added option to enable TLS on DataPrepperServer endpoints, on by…

wrijeff:
… default.

*Issue #, if available:* #321

*Description of changes:*
* Added `useTls`, `keyStoreFilePath` and `keyStorePassphrase` config parameters to `DataPrepperConfiguration.java`
  * TLS is enabled by default (and keyStore is empty, meaning apps will fail to start w/ default config), so I went back and passed custom configs explicitly turning off TLS to sample apps & integ tests
* Added unit tests for TLS-enabled DataPrepperServer
* Updated project setup page with new parameters

Had to change the port of `valid_data_prepper_config.yml` which is used in `DataPrepperTests.java`. At a high level, both `DataPrepperTests` and `DataPrepperServerTests` attempt to start an http server listening on a specific port. 

Previously this wasn't an issue because:
* `DataPrepperTests` did not pass in a config file, meaning it took the default port of 4900
* `DataPrepperServerTests` passed in a test config file, _all of which use port 1234_

Because we now need to pass in a config file to disable TLS, I started passing in one of the test config files to `DataPrepperTests`, which then hit "address in use" errors because now both test classes were attempting to use port 1234. I tried ensuring that each test class explicitly shut down its http server, but even then I still got bind errors.

We could improve the tests a bit by cleaning up the boundaries in each test (e.g. mocking HttpServer in tests which aren't explicitly testing Http capabilities, not relying on config files across multiple tests by mocking DataPrepperConfiguration, etc.) but this'll work for now.


data-prepper-core/src/integrationTest/resources/data_prepper.yml
```diff
@@ -0,0 +1 @@
+useTls: false
```
kowshikn, line 1: nit: For keeping it consistent with the source configuration and elasticsearch configs, can we call this as `ssl:false`

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -48,4 +65,42 @@ public void stop() {
...
+            keyStore.load(new FileInputStream(keystoreFilePath), passphraseArray);
+
+            KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance("SunX509");
+            keyManagerFactory.init(keyStore, passphraseArray);
+
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance("SunX509");
+            trustManagerFactory.init(keyStore);
+
+            sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(keyManagerFactory.getKeyManagers(), trustManagerFactory.getTrustManagers(), null);
```
kowshikn, line 90: The source uses `.crt`  and `.key`, would be great if we can follow the same pattern that way customers can have same experience. And JDK has moved towards `pkcs12` from exclusive `jks`.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -48,4 +65,42 @@ public void stop() {
...
+                                         final String keystoreFilePath,
+                                         final String passphrase) throws IOException {
+        final SSLContext sslContext;
+
+        try {
+            char[] passphraseArray = passphrase.toCharArray();
+            KeyStore keyStore = KeyStore.getInstance("JKS");
+            keyStore.load(new FileInputStream(keystoreFilePath), passphraseArray);
+
+            KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance("SunX509");
```
yadavcbala, line 83: similar comment - I recently stumbled across a similar issue with POC testing - `jks` is good and is default java key store but `pcks12` is language neutral - but also heard its little complex if we try to handle it using our own code [which we are not] - My experience with these is limited to just one or two use-cases so I may be ignoring a lot of other factors.

Also, the trust managers instance - SunX509 is vendor specific - we package one of our distribution with JDK so it should not a big issue but since we are hardcoding - it may result in NoSuchAlgorithmException if different vendor jdk is used. We may try a default algorithm like `TrustManagerFactory.getDefaultAlgorithm()` which i assume is `PKIX`. Again, I would leave it you and @kowshikn as my experience in these things is very limited.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -48,4 +65,42 @@ public void stop() {
...
+
+    private HttpServer createHttpsServer(final int port,
+                                         final String keystoreFilePath,
+                                         final String passphrase) throws IOException {
+        final SSLContext sslContext;
+
+        try {
+            char[] passphraseArray = passphrase.toCharArray();
+            KeyStore keyStore = KeyStore.getInstance("JKS");
+            keyStore.load(new FileInputStream(keystoreFilePath), passphraseArray);
```
yadavcbala, line 81: another nit: can we consider using `Files.newInputStream` over `new FileInputStream`

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -48,4 +65,42 @@ public void stop() {
...
+            KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance("SunX509");
+            keyManagerFactory.init(keyStore, passphraseArray);
+
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance("SunX509");
+            trustManagerFactory.init(keyStore);
+
+            sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(keyManagerFactory.getKeyManagers(), trustManagerFactory.getTrustManagers(), null);
+        } catch (Exception e) {
+            throw new IllegalStateException("Problem loading keystore to enable TLS", e);
```
yadavcbala, line 92: Question: Should we add a LOG.error here ? Not an issue as the calling code is only catching `IOException` but thinking, how the logs will be. I am fine either ways

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -48,4 +65,42 @@ public void stop() {
...
+        }
+
+        final HttpsServer server = HttpsServer.create(new InetSocketAddress(port), 0);
+        server.setHttpsConfigurator(new HttpsConfigurator(sslContext) {
+            public void configure(HttpsParameters params) {
+                SSLContext c = getSSLContext();
```
yadavcbala, line 98: nit: `sslContext`

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/PrometheusMetricsHandler.java
```diff
@@ -29,7 +29,7 @@ public void handle(HttpExchange exchange) throws IOException {
             exchange.sendResponseHeaders(HttpURLConnection.HTTP_OK, response.length);
             exchange.getResponseBody().write(response);
         } catch (Exception e) {
-            LOG.error("Encountered exception scarping prometheus meter registry", e);
+            LOG.error("Encountered exception scraping prometheus meter registry", e);
```
yadavcbala, line 32: good catch

docs/readme/project_setup.md
```diff
@@ -42,18 +42,24 @@ To run the example app against your local changes, use the docker found [here](h
 ## Data Prepper configuration
 Data Prepper allows the following properties to be configured
 
-* Server port
-* Log4J Properties
-  * Log level
-  * Log file
-  * Max file size
-  * Max backup index
+* `useTls`: boolean indicating TLS should be used for server APIs. Defaults to `true`
```
yadavcbala, line 45: looks much better with parameters

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
wrijeff: Good point regarding PKCS12 - should be easy to switch to that.

Regarding the `.crt` and `.key` files used by OTelSource/Armeria/Netty - they're actually following PKCS8 and that's accomplished by a lot of heavy lifting in `io.netty.handler.ssl` via [PemReader.java](https://github.com/whg333/netty-4.0.36.Final/blob/master/src/main/java/io/netty/handler/ssl/PemReader.java). 

We can do the same regex matching to strip the comment lines in the private key, loading that into a blank keystore initialized in memory, loading the certificate via CertificateFactory, etc. but it's quite a bit more effort than just loading a keystore file which already contains all that data. (I've been trying to implement the PKCS8 route but haven't quite got it working yet).

wrijeff:
Updated to use separate cert and private key files (PKCS 8). After going through all the trouble of setting up key and certificate loading, I'm inclined to go back and stick with supporting keystore files (.jks, .pem, .p12, etc.). 

Armeria uses key/cert files because that's what netty does:
* https://github.com/netty/netty/blob/ab8b8ae81b24439ff94085f3f4bd060c172c1712/handler/src/main/java/io/netty/handler/ssl/SslContextBuilder.java#L85-L94

They go through a lot of work to load TLS files, I'd rather stick with the built-in's like `KeyStore.getInstance()`. Worst case we can include instructions on how to create a keystore file given separate private key and cert files.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
wrijeff: The unhandled exception will bubble up in the logs, don't think we need to double log

wrijeff:
Follow-up: I've been thinking about @kowshikn's point about this change introducing a different way to enable TLS compared to what the OTel source uses (keystores vs private key/cert combos). 

I _think_ we should be able to pull keys and certs out of a keystore programmatically, so one potential solution would be to go back and have the OTel source take in a keystore too, from which we could pass the key/cert to Armeria like we currently do. Will poke around a bit more.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/SslUtil.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class SslUtil {
+
+    public static SSLContext createSslContext(final String keyStoreFilePath,
+                                              final String keyStorePassword,
+                                              final String privateKeyPassword) {
+        final SSLContext sslContext;
+
+        try {
+            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
```
yadavcbala, line 18: nit: final ?

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/SslUtil.java
```diff
@@ -0,0 +1,35 @@
...
+            keyStore.load(Files.newInputStream(Path.of(keyStoreFilePath)), keyStorePassword.toCharArray());
+
+            KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
```
yadavcbala, line 21: same as above

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/SslUtil.java
```diff
@@ -0,0 +1,35 @@
...
+            keyManagerFactory.init(keyStore, privateKeyPassword.toCharArray());
+
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
```
yadavcbala, line 24: same as above

yadavcbala (approved):
just non bloking nits

kowshikn (approved):

PR #347, opened 2021-02-22
Enabling dependabot on more directories, removing most of versionMap.

wrijeff:
*Issue #, if available:*

*Description of changes:*
* Removed version map (except for opentelemetry_proto for now)
* Enabled dependabot on remaining packages
* Included a lengthy comment as to why we're not using the latest micrometer release. 
  * Will open an issue on the micrometer repo asking why the maven artifact are showing as targeting Java 15 even though build.gradle specifies Java 8
    * Edit: PR to micrometer: https://github.com/micrometer-metrics/micrometer/pull/2471
  * Will also close/ignore dependabot upgrades for micrometer if they are opened until this is figured out

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,27 @@
+# Docs: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
 version: 2
 updates:
+- package-ecosystem: gradle
+  directory: "/data-prepper-api"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 3
+
+- package-ecosystem: gradle
+  directory: "/data-prepper-benchmarks"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 3
+
 - package-ecosystem: gradle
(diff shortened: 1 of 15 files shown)


.github/dependabot.yml
```diff
@@ -1,7 +1,27 @@
+# Docs: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates
 version: 2
 updates:
+- package-ecosystem: gradle
+  directory: "/data-prepper-api"
+  schedule:
+    interval: daily
```
chenqi0805, line 7: QUES: Could daily update be too frequent. Especially for API and core library.

chenqi0805 (approved):
LGTM. Just one question.

.github/dependabot.yml
wrijeff: Ah good point - I think we can change it to "weekly" once the initial round of dependency upgrades are complete. I'll create a follow-up PR later.

yadavcbala (approved):
LGTM

PR #355, opened 2021-02-23
dependabot.yml: explicitly listing out plugin directories.

wrijeff:
*Description of changes:*
* Noticed we weren't getting any pull requests for projects in the _plugins_ directory, so will attempt to fix by listing them all out.
* Also changing the schedule to "weekly" because we can trigger on-demand scans with https://github.com/opendistro-for-elasticsearch/data-prepper/network/updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,24 +4,54 @@ updates:
 - package-ecosystem: gradle
   directory: "/data-prepper-api"
   schedule:
-    interval: daily
-  open-pull-requests-limit: 3
-
-- package-ecosystem: gradle
-  directory: "/data-prepper-benchmarks"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 3
+    interval: weekly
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,24 +4,54 @@ updates:
...
-    interval: daily
-  open-pull-requests-limit: 3
+    interval: weekly
 
 - package-ecosystem: gradle
-  directory: "/data-prepper-benchmarks"
+  directory: "/data-prepper-core"
   schedule:
-    interval: daily
-  open-pull-requests-limit: 3
```
yadavcbala, line 14: Just for my information: what is the default limit ?

yadavcbala (approved):

.github/dependabot.yml
wrijeff: 5 is the default

chenqi0805 (approved):

Issue #320, opened 2021-02-10
Service-map plugin fails when worker > 1


kowshikn:
**Describe the bug**
Whenever we configure the service-map worker to be more than 1, we get service-map MapB create failure which causes the data-prepper to crash.
**To Reproduce**
1. Set `workers:` to any value greater than 1 and run the data-prepper

**Additional context**
***Logs***
```
180640 [service-map-pipeline-process-worker-2-thread-2] ERROR com.amazon.dataprepper.pipeline.ProcessWorker  ? Encountered exception during pipeline processing
org.mapdb.DBException$FileLocked: File is already opened and is locked: data/service-map/db-1612972144653
at org.mapdb.volume.Volume.lockFile(Volume.java:495)
at org.mapdb.volume.MappedFileVol.<init>(MappedFileVol.java:88)
at org.mapdb.volume.MappedFileVol$MappedFileFactory.factory(MappedFileVol.java:64)
at org.mapdb.volume.MappedFileVol$MappedFileFactory.makeVolume(MappedFileVol.java:38)
at org.mapdb.StoreDirect$volume$1.invoke(StoreDirect.kt:77)
at org.mapdb.StoreDirect$volume$1.invoke(StoreDirect.kt:16)
at org.mapdb.StoreDirect.<init>(StoreDirect.kt:76)
at org.mapdb.StoreDirect$Companion.make(StoreDirect.kt:57)
at org.mapdb.StoreDirect$Companion.make$default(StoreDirect.kt:56)
at org.mapdb.DBMaker$Maker.make(DBMaker.kt:450)
at com.amazon.dataprepper.plugins.processor.state.MapDbProcessorState.<init>(MapDbProcessorState.java:39)
at com.amazon.dataprepper.plugins.processor.ServiceMapStatefulProcessor.rotateWindows(ServiceMapStatefulProcessor.java:298)
at com.amazon.dataprepper.plugins.processor.ServiceMapStatefulProcessor.evaluateEdges(ServiceMapStatefulProcessor.java:169)
at com.amazon.dataprepper.plugins.processor.ServiceMapStatefulProcessor.execute(ServiceMapStatefulProcessor.java:98)
at com.amazon.dataprepper.pipeline.ProcessWorker.run(ProcessWorker.java:53)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[...]

PR #392, opened 2021-02-25
ENH: allow confine Prepper plugin instance to thread

chenqi0805:
*Issue #, if available:*
#320 

*Description of changes:*
This PR 

(1) Introduces a @SingleThread annotation for Prepper plugin to allow confinement of prepper instance to a single ProcessWorker.
(2) Modified `PrepperFactory` to return appropriately based on the @SingleThread.
(3) Modified `Pipeline` and `PipelineParser` correspondingly.
(4) Refactored some relevant test cases

### DIFF
--- a/data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/prepper/PrepperFactory.java
+++ b/data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/prepper/PrepperFactory.java
@@ -1,14 +1,28 @@
 package com.amazon.dataprepper.plugins.prepper;
 
+import com.amazon.dataprepper.model.annotations.SingleThread;
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.prepper.Prepper;
 import com.amazon.dataprepper.plugins.PluginFactory;
 import com.amazon.dataprepper.plugins.PluginRepository;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 @SuppressWarnings({"rawtypes"})
 public class PrepperFactory extends PluginFactory {
 
-    public static Prepper newPrepper(final PluginSetting pluginSetting) {
-        return (Prepper) newPlugin(pluginSetting, PluginRepository.getPrepperClass(pluginSetting.getName()));
+    public static List<Prepper> newPreppers(final PluginSetting pluginSetting) {
+        final Class<Prepper> clazz = PluginRepository.getPrepperClass(pluginSetting.getName());
+        if (clazz.isAnnotationPresent(SingleThread.class)) {
+            final List<Prepper> preppers = new ArrayList<>();
+            for (int i = 0; i < pluginSetting.getNumberOfProcessWorkers(); i++) {
+                preppers.add((Prepper) newPlugin(pluginSetting, clazz));
+            }
+            return preppers;
+        } else {
+            return Collections.singletonList((Prepper) newPlugin(pluginSetting, clazz));
+        }
     }
 }
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/model/annotations/SingleThread.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/model/annotations/SingleThread.java
@@ -1,0 +1,17 @@
+package com.amazon.dataprepper.model.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotates a Prepper plugin class of which an instance is required to be confined to a single thread.
+ */
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface SingleThread {
+}
--- a/data-prepper-core/build.gradle
+++ b/data-prepper-core/build.gradle
@@ -23,6 +23,7 @@ dependencies {
     implementation "org.reflections:reflections:0.9.12"
     implementation "io.micrometer:micrometer-core:1.6.4"
     implementation "io.micrometer:micrometer-registry-prometheus:1.6.4"
+    testCompile project(':data-prepper-plugins:common').sourceSets.test.output
     testImplementation "org.hamcrest:hamcrest:2.2"
     testImplementation "org.mockito:mockito-core:3.7.7"
 }
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
@@ -85,14 +85,9 @@ public class PipelineParser {
 
             LOG.info("Building preppers for the pipeline [{}]", pipelineName);
             final int prepperThreads = pipelineConfiguration.getWorkers();
-            final List<List<Prepper>> prepperSets = new ArrayList<>();
-            for (int i = 0; i < prepperThreads; i++) {
-                prepperSets.add(
-                        pipelineConfiguration.getPrepperPluginSettings().stream()
-                                .map(PrepperFactory::newPrepper)
-                                .collect(Collectors.toList())
-                );
-            }
+            final List<List<Prepper>> prepperSets = pipelineConfiguration.getPrepperPluginSettings().stream()
+                    .map(PrepperFactory::newPreppers)
+                    .collect(Collectors.toList());
             final int readBatchDelay = pipelineConfiguration.getReadBatchDelay();
 
             LOG.info("Building sinks for the pipeline [{}]", pipelineName);
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
@@ -21,6 +21,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -83,10 +84,15 @@ public class PipelineParser {
             final Buffer buffer = BufferFactory.newBuffer(pipelineConfiguration.getBufferPluginSetting());
 
             LOG.info("Building preppers for the pipeline [{}]", pipelineName);
-            final List<Prepper> preppers = pipelineConfiguration.getPrepperPluginSettings().stream()
-                    .map(PrepperFactory::newPrepper)
-                    .collect(Collectors.toList());
             final int prepperThreads = pipelineConfiguration.getWorkers();
+            final List<List<Prepper>> prepperSets = new ArrayList<>();
+            for (int i = 0; i < prepperThreads; i++) {
+                prepperSets.add(
+                        pipelineConfiguration.getPrepperPluginSettings().stream()
+                                .map(PrepperFactory::newPrepper)
+                                .collect(Collectors.toList())
+                );
+            }
             final int readBatchDelay = pipelineConfiguration.getReadBatchDelay();
 
             LOG.info("Building sinks for the pipeline [{}]", pipelineName);
@@ -94,7 +100,7 @@ public class PipelineParser {
                     .map(this::buildSinkOrConnector)
                     .collect(Collectors.toList());
 
-            final Pipeline pipeline = new Pipeline(pipelineName, source, buffer, preppers, sinks, prepperThreads, readBatchDelay);
+            final Pipeline pipeline = new Pipeline(pipelineName, source, buffer, prepperSets, sinks, prepperThreads, readBatchDelay);
             pipelineMap.put(pipelineName, pipeline);
         } catch (Exception ex) {
             //If pipeline construction errors out, we will skip that pipeline and proceed
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
@@ -53,7 +53,8 @@ public class Pipeline {
      * @param name                     name of the pipeline
      * @param source                   source from where the pipeline reads the records
      * @param buffer                   buffer for the source to queue records
-     * @param prepperSets               prepper that is applied to records
+     * @param prepperSets               prepper sets that will be applied to records. Each set includes either a single shared prepper instance
+     *                                  or multiple instances with each to be accessed only by a single {@link ProcessWorker}.
      * @param sinks                    sink to which the transformed records are posted
      * @param prepperThreads         configured or default threads to parallelize prepper work
      * @param readBatchTimeoutInMillis configured or default timeout for reading batch of records from buffer
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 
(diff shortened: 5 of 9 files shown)


data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/prepper/PrepperFactory.java
```diff
@@ -1,14 +1,28 @@
...
+        final Class<Prepper> clazz = PluginRepository.getPrepperClass(pluginSetting.getName());
+        if (clazz.isAnnotationPresent(SingleThread.class)) {
+            final List<Prepper> preppers = new ArrayList<>();
+            for (int i = 0; i < pluginSetting.getNumberOfProcessWorkers(); i++) {
+                preppers.add((Prepper) newPlugin(pluginSetting, clazz));
+            }
+            return preppers;
+        } else {
+            return Collections.singletonList((Prepper) newPlugin(pluginSetting, clazz));
+        }
```
wrijeff, line 26: Should the conditions be flipped here? If annotated, return singletonList?

wrijeff (commented):
A few questions/comments about these changes:

1. By introducing a single thread limit on specific components, won't we be introducing an immediate bottleneck on performance? Example: OTelSource has 16 workers, RawPrepper has 16 workers, ServiceMap has 1 worker. Now the total throughput of the Data Prepper instance is limited by the single-threaded component? And if that's the case, why not just make everything single-threaded?
2. Can't ServiceMapStatefulPrepper manage this race condition by putting a simple lock on the `evaluateEdges()` function? If the lock is already acquired (`tryLock()` returns false), just continue processing as normal. 
    * It looks like there is already some `synchronized` code in the prepper, maybe we just didn't make it completely thread safe. (See `isMasterInstance()`, every thread is _not_ intended to rotate the window).

My personal preference would be to manage parallelism in each component for the reasons above. In this case we'd just need to fix the thread safety issue in the prepper, which looks to be almost complete.

chenqi0805:
@wrijeff Thanks for the questions and comments. Below are the explanations.

> 1. By introducing a single thread limit on specific components, won't we be introducing an immediate bottleneck on performance? Example: OTelSource has 16 workers, RawPrepper has 16 workers, ServiceMap has 1 worker. Now the total throughput of the Data Prepper instance is limited by the single-threaded component? And if that's the case, why not just make everything single-threaded?

The introduction of @SingleThread annotation is to produce the corresponding number of duplicate prepper instances according to number of workers and then pass them into the worker in 1-to-1 correspondence (e.g. 16 ServiceMapSateful instances passed into 16 workers respectively). So it should not have impaired the parallel processing. Instead, the motivation is to provide developer an additional option to deal with concurrency and parallelism(i.e. a prepper instance is owned by a thread/worker). E.g. Even if a stateful prepper is not threadsafe, one could still let it run in parallel by adding this annotation w/o worry about concurrency issue.

Note: Maybe the annotation name @SingleThread is a little confusing. My initial thought is to use @PerThread but might still not be manifest.

> 2. Can't ServiceMapStatefulPrepper manage this race condition by putting a simple lock on the evaluateEdges() function? If the lock is already acquired (tryLock() returns false), just continue processing as normal.
It looks like there is already some synchronized code in the prepper, maybe we just didn't make it completely thread safe. (See isMasterInstance(), every thread is not intended to rotate the window).

Yeah. That idea will also work under the current expectation(i.e. only one instance of a prepper type is created and is shared across all workers). But as mentioned above, the idea is to provide additional options for dealing with parallelism in stateful prepper development. 

> My personal preference would be to manage parallelism in each component for the reasons above. In this case we'd just need to fix the thread safety issue in the prepper, which looks to be almost complete.

This is closer to approach 2 as in #320. But we think introduce this additional option could benefit future custom prepper plugin contribution.

data-prepper-plugins/common/src/main/java/com/amazon/dataprepper/plugins/prepper/PrepperFactory.java
chenqi0805: See the explanation on what the annotation means. The name of the annotation class might not be manifest. Thoughts are welcome

wrijeff:
Thanks @chenqi0805 for the reply, I misinterpreted the meaning of the `@SingleThread` annotation. More follow-up questions :)

1. Do we know if this change would address the root cause of #320? From the description it looks like the issue was caused by multiple threads all attempting to access a single, locked DB file (I haven't dug in for more detail yet). How would switching from a single Java object shared by X threads, to instead having X copies of that object solve this issue?
2. What actually was the root cause of #320? As I mentioned previously it appears that we knew about multi-threaded issues and [attempted to code around them](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/master/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java#L364), so what was missed?

General comment: I'm _very_ wary of adding structural features that we "might" use in the future. They add complexity, and currently we don't have a framework setup to measure the performance impact of such a change.

chenqi0805:
@wrijeff 

> 1. Do we know if this change would address the root cause of #320? From the description it looks like the issue was caused by multiple threads all attempting to access a single, locked DB file (I haven't dug in for more detail yet). How would switching from a single Java object shared by X threads, to instead having X copies of that object solve this issue?

> 2. What actually was the root cause of #320? As I mentioned previously it appears that we knew about multi-threaded issues and attempted to code around them, so what was missed?

The root cause is that all workers share the master instance and they are racing to rotate the window due to misalignment between the ownership assumption of ServiceMapStateful and data-prepper-api. 

However, the exception stacktrace in #320 that appeared in load test is hard to reproduce in e2e test. My speculation is that e2e test does not have the throughput that led to two threads race to create the db file under duplicate name (db file is suffixed by time in milli second which is the resolution). But that stacktrace is not the root cause to fix. At least we need to fix the misalignment first and see if there is any other concurrency issue.

chenqi0805:
@wrijeff Having tested the branch on trace-analytics-sample-app. The two tracebacks that appear in the current main branch:

1. `180640 [service-map-pipeline-process-worker-2-thread-2] ERROR com.amazon.dataprepper.pipeline.ProcessWorker  ? Encountered exception during pipeline processing
org.mapdb.DBException$FileLocked: File is already opened and is locked: data/service-map/db-1612972144653`

2.

```
2021-03-03T17:05:07,192 [service-map-pipeline-process-worker-2-thread-1] ERROR com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper - Caught exception trying to get trace group name
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
        at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64) ~[?:?]
        at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70) ~[?:?]
        at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248) ~[?:?]
        at java.util.Objects.checkIndex(Objects.java:359) ~[?:?]
        at java.util.ArrayList.get(ArrayList.java:427) ~[?:?]
        at java.util.Collections$SynchronizedList.get(Collections.java:2460) ~[?:?]
        at org.mapdb.WriteAheadLog.walGetRecord(WriteAheadLog.java:863) ~[data-prepper.jar:0.7.0-alpha]
        at org.mapdb.StoreWAL.get(StoreWAL.kt:522) ~[data-prepper.jar:0.7.0-alpha]
        at org.mapdb.BTreeMap.getNode(BTreeMap.kt:800) ~[data-prepper.jar:0.7.0-alpha]
        at org.mapdb.BTreeMap.getNonBinary(BTreeMap.kt:271) ~[data-prepper.jar:0.7.0-alpha]
        at org.mapdb.BTreeMap.get(BTreeMap.kt:235) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.state.MapDbPrepperState.get(MapDbPrepperState.java:61) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper.getTraceGroupName(ServiceMapStatefulPrepper.java:254) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper.lookupParentSpan(ServiceMapStatefulPrepper.java:218) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper.access$000(ServiceMapStatefulPrepper.java:32) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper$1.apply(ServiceMapStatefulPrepper.java:211) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper$1.apply(ServiceMapStatefulPrepper.java:208) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.state.MapDbPrepperState.lambda$iterate$1(MapDbPrepperState.java:85) ~[data-prepper.jar:0.7.0-alpha]
        at java.util.Iterator.forEachRemaining(Iterator.java:133) ~[?:?]
        at com.amazon.dataprepper.plugins.prepper.state.MapDbPrepperState.iterate(MapDbPrepperState.java:84) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper.evaluateEdges(ServiceMapStatefulPrepper.java:164) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.plugins.prepper.ServiceMapStatefulPrepper.doExecute(ServiceMapStatefulPrepper.java:114) ~[data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.model.prepper.AbstractPrepper.lambda$execute$0(AbstractPrepper.java:39) ~[data-prepper.jar:0.7.0-alpha]
        at io.micrometer.core.instrument.composite.CompositeTimer.record(CompositeTimer.java:57) [data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.model.prepper.AbstractPrepper.execute(AbstractPrepper.java:39) [data-prepper.jar:0.7.0-alpha]
        at com.amazon.dataprepper.pipeline.ProcessWorker.run(ProcessWorker.java:57) [data-prepper.jar:0.7.0-alpha]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
        at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630) [?:?]
        at java.lang.Thread.run(Thread.java:832) [?:?]

```

has not been detected with the build from current branch. However I found the above two tracebacks are flaky which means sometimes I have to restart data-prepper or sample-app a few times to reproduce. So it is not guaranteed they are completely eliminated with the current fix. Need to keep watching monitoring in the future.

wrijeff:
Thanks @chenqi0805 for reproducing the issue and validating the fix. 

I'm still of the opinion that this change is potentially overkill for this bug, and would prefer to just fix the thread safety of the ServiceMapStatefulPrepper instead of introducing a new single/multi-threading paradigm. Especially without having the ability to benchmark the performance impact of such a change.

For reference, here's a diff which adds a simple Java lock to the ServiceMapStatefulPrepper and appears to mitigate the issue (tested via docker-compose):
```
diff --git a/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java b/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
index d8ac1e9..e1057e8 100644
--- a/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
+++ b/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
@@ -24,6 +24,8 @@ import java.util.Objects;
 import java.util.TreeMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -56,6 +58,8 @@ public class ServiceMapStatefulPrepper extends AbstractPrepper<Record<ExportTrac

     private final int thisPrepperId;

+    private final Lock lock;
+
     public ServiceMapStatefulPrepper(final PluginSetting pluginSetting) {
         this(pluginSetting.getIntegerOrDefault(ServiceMapPrepperConfig.WINDOW_DURATION, ServiceMapPrepperConfig.DEFAULT_WINDOW_DURATION) * TO_MILLIS,
                 new File(ServiceMapPrepperConfig.DEFAULT_LMDB_PATH),
@@ -70,6 +74,7 @@ public class ServiceMapStatefulPrepper extends AbstractPrepper<Record<ExportTrac
         super(pluginSetting);

         ServiceMapStatefulPrepper.clock = clock;
+        this.lock = new ReentrantLock();
         this.thisPrepperId = preppersCreated.getAndIncrement();
         if (isMasterInstance()) {
             previousTimestamp = ServiceMapStatefulPrepper.clock.millis();
@@ -159,6 +164,10 @@ public class ServiceMapStatefulPrepper extends AbstractPrepper<Record<ExportTrac
      * @return Set of Record<String> containing json representation of ServiceMapRelationships found
      */
     private Collection<Record<String>> evaluateEdges() {
+        if (!lock.tryLock()) {
+            return EMPTY_COLLECTION;
+        }
+
         try {
             final Stream<ServiceMapRelationship> previousStream = previousWindow.iterate(relationshipIterationFunction, preppersCreated.get(), thisPrepperId).stream().flatMap(serviceMapEdgeStream -> serviceMapEdgeStream);
             final Stream<ServiceMapRelationship> currentStream = currentWindow.iterate(relationshipIterationFunction, preppersCreated.get(), thisPrepperId).stream().flatMap(serviceMapEdgeStream -> serviceMapEdgeStream);
@@ -192,6 +201,8 @@ public class ServiceMapStatefulPrepper extends AbstractPrepper<Record<ExportTrac
             return serviceDependencyRecords;
         } catch (InterruptedException e) {
             throw new RuntimeException(e);
+        } finally {
+            lock.unlock();
         }
     }
```
We can clean up the Prepper a bit more in future PRs (moving away from static volatile, removing need for "master" instance, etc.), but for now this addresses the race condition described in the original issue.

chenqi0805:
> I'm still of the opinion that this change is potentially overkill for this bug, and would prefer to just fix the thread safety of the ServiceMapStatefulPrepper instead of introducing a new single/multi-threading paradigm.

Understood. Personally I am OK with that approach. But since the current approach has been synced with @yadavcbala @kowshikn @AustinTag . Any input or thoughts?

> Especially without having the ability to benchmark the performance impact of such a change.

I am not sure what this means specifically b/c my understanding of performance benchmark is always treat data-prepper as a black box.

Note: Have not dived too much into the shared thread ownership implementation of ServiceMapStateful but I think we should remove `isMasterInstance` logic since it is no longer meaningful.

wrijeff:
> I am not sure what this means specifically b/c my understanding of performance benchmark is always treat data-prepper as a black box.

Ah, I meant that we're not yet measuring throughput (benchmarking) in an automated fashion, and I would be interested in how this change might affect performance (it could also be a positive change).

> Note: Have not dived too much into the shared thread ownership implementation of ServiceMapStateful but I think we should remove `isMasterInstance` logic since it is no longer meaningful.

Agreed. I'm wondering if the original intent was to have many single-threaded objects but something got changed in the API? Either way, once we straighten things out it'd be good to document indicating if components are expected to be thread-safe or not.

wrijeff (approved):
Discussed offline, approving as this provides immediate value for the next release. However I would like to consider eventually moving back to requiring all pipeline components to be thread safe (after reviewing performance benchmarks).

What happened here is we had a component implemented to be run as multiple, single-threaded objects. This was done by shared, static volatile variables and countdown latches. The actual DP pipeline was implemented to create single-copy, multi-threaded objects. So the two approaches to fix were to either make the component thread-safe, or make the pipeline support multi-objects.

chenqi0805:
@wrijeff Thanks for the deep diving into the service-map. I will ping others for a second approval.

yadavcbala (approved):

PR #393, opened 2021-02-26
Switching to log4j2.

wrijeff:
*Description of changes:*
* Switching the logging backend from log4j to log4j2
* Had to include a new log4j2.properties file as the upgrade isn't 100% backwards compatible

I also had a question as to if we should keep the log4j runtime configuration stuff via DP config, as it seems we're already passing in a log4j.properties file as part of the startup command - figured we can discuss that here.

I'll go back and clean up some of the docs around the old log4j stuff once this is merged. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ subprojects {
         implementation "com.google.guava:guava:29.0-jre"
         implementation "org.apache.logging.log4j:log4j-core:2.11.1"
         implementation "org.slf4j:slf4j-api:1.7.30"
-        implementation "org.slf4j:slf4j-log4j12:1.7.30"
+        implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.14.0"
         testImplementation("junit:junit:4.13") {
             exclude group: 'org.hamcrest' // workaround for jarHell
         }
@@ -25,6 +25,8 @@ subprojects {
     jacocoTestReport {
         dependsOn test // tests are required to run before generating the report
     }
+
+    task allDeps(type: DependencyReportTask) {}
 }
 
 configure(coreProjects) {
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -16,7 +16,7 @@ subprojects {
         implementation "com.google.guava:guava:29.0-jre"
         implementation "org.apache.logging.log4j:log4j-core:2.11.1"
         implementation "org.slf4j:slf4j-api:1.7.30"
-        implementation "org.slf4j:slf4j-log4j12:1.7.30"
+        implementation "org.apache.logging.log4j:log4j-slf4j-impl:2.14.0"
```
AustinTag, line 19: Nit: The "2.14.0" version is re-used in a few places. Seems that it can be added to the version map

AustinTag (approved):
Approving, on the condition that we manually tested that the log level configuration still works. For now, the only way we have of verifying the log level configuration is by setting the log level via config, and looking at the log file or console to see that the logging is correct. We can chat offline about how to do that if you like

wrijeff:
> Approving, on the condition that we manually tested that the log level configuration still works. For now, the only way we have of verifying the log level configuration is by setting the log level via config, and looking at the log file or console to see that the logging is correct. We can chat offline about how to do that if you like

Discussed offline, tested the log4j2.properties file to adjust log thresholds. Will likely remove the "log configuration via data-prepper.yml feature" as users can already just pass in their modified .properties file. WIll be a separate PR with documentation cleanup.

build.gradle
wrijeff: Removed most of the version map in #347 in favor of automated dependency management (just FYI)

yadavcbala (approved):

PR #438, opened 2021-03-17
ENH: introduce cache for traceId to traceGroup in OtelTraceRawPrepper

chenqi0805:
*Issue #, if available:*

*Description of changes:*
This PR introduces Guava Cache for storing traceId to traceGroup mapping so that the traceGroup is always looked up before flushing any batch of spans.

Note: For persistent storage, the expiration function of MapDB seems to be broken(https://github.com/jankotek/mapdb/issues/975. Or maybe I misunderstood how it should be used). So I used Guava cache with bounded size and ttl.

-----

    indicated in the file; or

    in the file; or

    it.


data-prepper-plugins/otel-trace-raw-prepper/build.gradle
```diff
@@ -13,6 +13,7 @@ dependencies {
     implementation "com.linecorp.armeria:armeria-grpc:1.5.0"
     implementation "com.fasterxml.jackson.core:jackson-databind:2.12.2"
     implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.2"
+    implementation group: 'com.google.guava', name: 'guava', version: '11.0.2'
```
wrijeff, line 16: Can we use a more recent version? - https://mvnrepository.com/artifact/com.google.guava/guava

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
```diff
@@ -5,10 +5,12 @@
 import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.prepper.AbstractPrepper;
 import com.amazon.dataprepper.model.record.Record;
+import com.amazon.dataprepper.plugins.prepper.oteltrace.model.MapDBTraceIdTraceGroupCache;
```
wrijeff: Does this still exist?

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/TraceIdTraceGroupCache.java
```diff
@@ -0,0 +1,35 @@
+package com.amazon.dataprepper.plugins.prepper.oteltrace.model;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+public class TraceIdTraceGroupCache {
```
wrijeff, line 8: Do we need a separate class for this (instead of just using the Guava cache directly)?

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OtelTraceRawPrepperConfig.java
```diff
@@ -5,4 +5,6 @@
     static final long DEFAULT_TG_FLUSH_INTERVAL = 180L;
     static final String ROOT_SPAN_FLUSH_DELAY = "root_span_flush_delay";
     static final long DEFAULT_ROOT_SPAN_FLUSH_DELAY = 30L;
+    static final long DEFAULT_TRACE_ID_TTL = 300L;
```
wrijeff, line 8: nit - Can we get the units included in the variable names? e.g. `_TTL_SECONDS`, `_DELAY_SECONDS`

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
```diff
@@ -197,13 +209,22 @@ private void processDescendantRawSpan(final RawSpan rawSpan) {
...
-                                log.warn("Missing root span for SpanId: {}", rawSpan.getSpanId());
+                            if (traceGroup != null) {
+                                rawSpans.forEach(rawSpan -> {
+                                    rawSpan.setTraceGroup(traceGroup);
+                                    recordsToFlush.add(rawSpan);
+                                });
+                            } else {
+                                rawSpans.forEach(rawSpan -> {
+                                    recordsToFlush.add(rawSpan);
+                                    log.warn("Missing root span for SpanId: {}", rawSpan.getSpanId());
```
wrijeff, line 225: Might be more accurate to say "missing trace group for SpanId" (if there's ever a case where we have the root span yet trace group is still somehow null).

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
```diff
@@ -122,6 +131,9 @@ private void processRawSpan(final RawSpan rawSpan) {
     }
 
     private void processRootRawSpan(final RawSpan rawSpan) {
+        // TODO: flush descendants here to get rid of DelayQueue
+        // TODO: safe-guard against null traceGroup for rootSpan?
+        traceIdTraceGroupCache.put(rawSpan.getTraceId(), rawSpan.getTraceGroup());
```
wrijeff, line 135: Just curious - does putting an item in the cache that already exists extend its TTL?

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/TraceIdTraceGroupCache.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final Cache<String, String> map;
+
+    public TraceIdTraceGroupCache(final int concurrencyLevel, final long maximumSize, final long ttl) {
```
wrijeff, line 12: nit - Would recommend `ttlSeconds` as arg name

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
chenqi0805: Yes. According to the docs of expireAfterWrite: 
```
Specifies that each entry should be automatically removed from the cache once a fixed duration
has elapsed after the entry's creation, or the most recent replacement of its value.
```

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
chenqi0805: This is a mis. Fixed

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/TraceIdTraceGroupCache.java
chenqi0805: We could. I separate this class for (1) testability (2) we might still replace Guava cache in the future and it might incur more configuration parameters.

wrijeff: Gotcha. Then for now I'd suggest just using the cache directly without a wrapper. They [already test extensively](https://github.com/google/guava/blob/master/guava-tests/test/com/google/common/cache/LocalCacheTest.java), and  currently there isn't any added functionality gained at the cost of more abstraction that I can see.

Definitely agree on the advantage of being able to swap out the backing cache later. If we want to do that then we'll just need to swap out the 7 lines of code in OTelTraceRawPrepper to get back to using a wrapper of our own.

That being said, if still you prefer to keep this wrapper then consider changing the name to make it less coupled to this specific usecase. Could use generics to let the caller decide what the cache will store instead of hardcoding the <String, String> implementation too.

wrijeff (approved):
👍

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
```diff
@@ -66,11 +70,17 @@
...
+                OtelTraceRawPrepperConfig.TRACE_FLUSH_INTERVAL, OtelTraceRawPrepperConfig.DEFAULT_TG_FLUSH_INTERVAL_SEC);
         rootSpanFlushDelay = SEC_TO_MILLIS * pluginSetting.getLongOrDefault(
-                OtelTraceRawPrepperConfig.ROOT_SPAN_FLUSH_DELAY, OtelTraceRawPrepperConfig.DEFAULT_ROOT_SPAN_FLUSH_DELAY);
+                OtelTraceRawPrepperConfig.ROOT_SPAN_FLUSH_DELAY, OtelTraceRawPrepperConfig.DEFAULT_ROOT_SPAN_FLUSH_DELAY_SEC);
         Preconditions.checkArgument(rootSpanFlushDelay <= traceFlushInterval,
                 "rootSpanSetFlushDelay should not be greater than traceFlushInterval.");
+        final int numProcessWorkers = pluginSetting.getNumberOfProcessWorkers();
+        traceIdTraceGroupCache = CacheBuilder.newBuilder()
+                .concurrencyLevel(numProcessWorkers)
+                .maximumSize(OtelTraceRawPrepperConfig.MAX_TRACE_ID_CACHE_SIZE_PER_THREAD * numProcessWorkers)
```
kowshikn, line 81: I don't understand this calculation? What's the idea here?

chenqi0805: This is just to make it more precise in max size control. According to the doc of this setting:

```
/**
   * Specifies the maximum number of entries the cache may contain.
   *
   * <p>Note that the cache <b>may evict an entry before this limit is exceeded</b>. For example, in
   * the current implementation, when {@code concurrencyLevel} is greater than {@code 1}, each
   * resulting segment inside the cache <i>independently</i> limits its own size to approximately
   * {@code maximumSize / concurrencyLevel}.
```

chenqi0805: Set to a fixed max size for the entire pipeline

kowshikn: As discussed offline, we need to set the cache size based on for entire pipeline rather based on threads. I would recommend a single property.

kowshikn (approved):

PR #493, opened 2021-04-08
ENH: expanding traceGroup fields

chenqi0805:
*Issue #, if available:*

*Description of changes:*

-----
This PR 

1. expands traceGroup fields to include

- name
- endTime
- durationInNanos
- statusCode

in the following plugins

- otel-trace-raw-prepper
- otel-trace-group-prepper

2. Adapted test suites.

    indicated in the file; or

    in the file; or

    it.


data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -126,13 +148,24 @@ private void sendExportTraceServiceRequestToSource(final int port, final ExportT
...
         return sources;
     }
 
     public static ResourceSpans getResourceSpans(final String serviceName, final String spanName, final byte[]
-            spanId, final byte[] parentId, final byte[] traceId, final Span.SpanKind spanKind) {
+            spanId, final byte[] parentId, final byte[] traceId, final Span.SpanKind spanKind, final String endTime,
+                                                 final Long durationInNanos, final Integer statusCode) {
         final ByteString parentSpanId = parentId != null ? ByteString.copyFrom(parentId) : ByteString.EMPTY;
+        final long endTimeInNanos = convertTimeStampToNanos(endTime);
+        final long startTimeInNanos = new BigDecimal(endTimeInNanos).subtract(new BigDecimal(durationInNanos)).longValue();
```
wrijeff, line 168: Can we not just subtract longs directly without needing a conversion?

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/OTelTraceGroupPrepper.java
```diff
@@ -83,19 +85,20 @@ public OTelTraceGroupPrepper(final PluginSetting pluginSetting) {
...
+        final Map<String, TraceGroupWrapper> traceIdToTraceGroup = searchTraceGroupByTraceIds(traceIdsToLookUp);
         for (final Map.Entry<Record<String>, Map<String, Object>> entry: recordMissingTraceGroupToRawSpanMap.entrySet()) {
             final Record<String> record = entry.getKey();
             final Map<String, Object> rawSpanMap = entry.getValue();
             final String traceId = (String) rawSpanMap.get(OTelTraceGroupPrepperConfig.TRACE_ID_FIELD);
-            final String traceGroup = traceIdToTraceGroup.get(traceId);
-            if (!Strings.isNullOrEmpty(traceGroup)) {
-                rawSpanMap.put(OTelTraceGroupPrepperConfig.TRACE_GROUP_FIELD, traceGroup);
+            final TraceGroupWrapper traceGroup = traceIdToTraceGroup.get(traceId);
+            if (Objects.nonNull(traceGroup)) {
```
wrijeff, line 94: Super nit - this kind of threw me off instead of the usual `!= null`.

Might suggest only using the Objects methods when required (like in lambda functions) to keep consistent with the other uses of `!= null`, of which there are 100+.

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/OTelTraceGroupPrepper.java
```diff
@@ -83,19 +85,20 @@ public OTelTraceGroupPrepper(final PluginSetting pluginSetting) {
...
                 try {
+                    Map<String, Object> traceGroupMap = OBJECT_MAPPER.convertValue(traceGroup, MAP_TYPE_REFERENCE);
+                    rawSpanMap.putAll(traceGroupMap);
                     final String newData = OBJECT_MAPPER.writeValueAsString(rawSpanMap);
                     recordsOut.add(new Record<>(newData, record.getMetadata()));
                     recordsOutFixedTraceGroupCounter.increment();
-                } catch (JsonProcessingException e) {
+                } catch (Exception e) {
                     recordsOut.add(record);
```
wrijeff, line 102: Similar to the RawPrepper exception handling, do we want to instead drop the record and emit a metric instead of adding it to the output records?

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/OTelTraceGroupPrepper.java
```diff
@@ -110,20 +113,17 @@ public OTelTraceGroupPrepper(final PluginSetting pluginSetting) {
...
             final SearchHit[] searchHits = searchResponse.getHits().getHits();
             Arrays.asList(searchHits).forEach(searchHit -> {
-                final DocumentField traceIdDocField = searchHit.field(OTelTraceGroupPrepperConfig.TRACE_ID_FIELD);
-                final DocumentField traceGroupDocField = searchHit.field(OTelTraceGroupPrepperConfig.TRACE_GROUP_FIELD);
-                if (Stream.of(traceIdDocField, traceGroupDocField).allMatch(Objects::nonNull)) {
-                    final String traceId = searchHit.field(OTelTraceGroupPrepperConfig.TRACE_ID_FIELD).getValue();
-                    final String traceGroup = searchHit.field(OTelTraceGroupPrepperConfig.TRACE_GROUP_FIELD).getValue();
-                    traceIdToTraceGroup.put(traceId, traceGroup);
+                Map.Entry<String, TraceGroupWrapper> entry = fromSearchHitToMapEntry(searchHit);
+                if (Objects.nonNull(entry)) {
```
wrijeff, line 125: Same comment about `!= null`. However in this case I suggest an optional a bit further down, so the statement would instead be `if (entry.isPresent()) {`

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/TraceGroupWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+        return endTime;
+    }
+
+    public Long getDurationInNanos() {
+        return durationInNanos;
+    }
+
+    public Integer getStatusCode() {
+        return statusCode;
+    }
```
wrijeff, line 39: Suggest always using primitives over objects when possible (long, int)

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/RawSpan.java
```diff
@@ -183,7 +185,8 @@ public void setTraceGroup(String traceGroup) {
         return attributes;
     }
 
-    public String getTraceGroup() {
+    @JsonUnwrapped(prefix = "traceGroup.")
```
wrijeff, line 188: Just curious - what's this for? I haven't used many of the JSON annotations previously.

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/TraceGroup.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.opentelemetry.proto.trace.v1.Span;
+
+import java.util.Objects;
+
+public class TraceGroup {
+    private final String name;
+    private final String endTime;
+    private final Integer statusCode;
+    private final Long durationInNanos;
```
wrijeff, line 11: Similar comment about preferring primitives over objects when available.

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/model/RawSpan.java
chenqi0805: Oh. This is to flatten out the nested traceGroup attributes by adding the prefix. So that we have like traceGroup.name under raw span json.

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/TraceGroupWrapper.java
chenqi0805: The reason I did not use primitives is to reserve for null as otherwise in case the trace group has not been populated, the default value of int and long will be confused with the meaningful value range.

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/OTelTraceGroupPrepper.java
chenqi0805: Good catch. I will emit the metric but probably not drop the record here b/c the exception is associated with the `newData` (Thus we can still output the old record).

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
chenqi0805: Good catch. This is an overkill

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -164,6 +200,13 @@ public static ExportTraceServiceRequest getExportTraceServiceRequest(List<Resour
                 .build();
     }
 
+    private static long convertTimeStampToNanos(String time) {
+        Instant instant = Instant.parse(time);
+        BigDecimal nanos = new BigDecimal(instant.getNano());
+        BigDecimal epochNanoSeconds = new BigDecimal(instant.getEpochSecond()).multiply(SEC_TO_NANOS).add(nanos);
+        return epochNanoSeconds.longValue();
```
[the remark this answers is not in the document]
chenqi0805: Will give a try.

data-prepper-plugins/otel-trace-group-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltracegroup/TraceGroupWrapper.java
wrijeff: I see your point, that's an annoying detail. I could see durationInNanos ever being `0` as unlikely, but statusCode is often `0` 😐

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -35,12 +42,28 @@
 import static org.awaitility.Awaitility.await;
 
 public class EndToEndRawSpanTest {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<Map<String, Object>>() {};
+    private static final BigDecimal SEC_TO_NANOS = new BigDecimal(1_000_000_000);
```
wrijeff, line 47: No longer used

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -24,6 +28,9 @@
 import org.junit.Test;
 
 import java.io.IOException;
+import java.math.BigDecimal;
```
wrijeff, line 31: No longer used

wrijeff (approved):
👍

data-prepper-plugins/elasticsearch/src/main/resources/otel-v1-apm-span-index-template.json
```diff
@@ -40,8 +40,21 @@
         "type": "keyword"
       },
       "traceGroup": {
-        "ignore_above": 1024,
-        "type": "keyword"
+        "properties": {
```
kowshikn, line 43: This will break existing index data and kibana plugin. Did we test it?

wrijeff: Good question - just built and tested, ran into this:

>2021-04-14T20:06:55,580 [raw-pipeline-process-worker-3-thread-2] WARN  com.amazon.dataprepper.plugins.sink.elasticsearch.ElasticsearchSink - Document [index {[otel-v1-apm-span][_doc][5cc8fdea8bda115f], source[{"traceId":"44bbcc035e9b60f4b321d75d341e3ebe","spanId":"5cc8fdea8bda115f","traceState":"","parentSpanId":"1d5fe371e312dbd3","name":"HTTP POST","kind":"SPAN_KIND_CLIENT","startTime":"2021-04-14T20:06:20.896851712Z","endTime":"2021-04-14T20:06:20.901840128Z","durationInNanos":4988416,"serviceName":"order","events":[],"links":[],"droppedAttributesCount":0,"droppedEventsCount":0,"droppedLinksCount":0,"traceGroup.name":"client_delivery_status","traceGroup.endTime":"2021-04-14T20:06:20.913838336Z","traceGroup.statusCode":0,"traceGroup.durationInNanos":59541504,"span.attributes.http@url":"http://localhost:8087/logs","instrumentationLibrary.version":"0.14b0","resource.attributes.telemetry@sdk@language":"python","span.attributes.http@status_text":"","resource.attributes.telemetry@sdk@version":"0.14b0","resource.attributes.service@instance@id":"140366117791112","resource.attributes.service@name":"order","span.attributes.component":"http","status.code":0,"instrumentationLibrary.name":"opentelemetry.instrumentation.requests","span.attributes.http@method":"POST","resource.attributes.telemetry@sdk@name":"opentelemetry","resource.attributes.host@hostname":"xxxxxx.ec2.internal","span.attributes.http@status_code":200}]}] has failure: ElasticsearchException[Elasticsearch exception [type=mapper_parsing_exception, reason=Could not dynamically add mapping for field [traceGroup.name]. Existing mapping for [traceGroup] must be of type object but found [keyword].]]

The data never reaches the cluster, it fails to be written by the ES sink. So this is a breaking change - any solutions to this other than requiring users to clear their data?

kowshikn: This is because the ES index template has changed. Earlier TraceGroup was keyword, now its not. This is why we should maybe consider them as separate fields.

chenqi0805: Agreed. That may be the only option to support backward compatibility. Maybe we can do field names as such
```
traceGroup (name)
traceGroup@statusCode
traceGroup@endTime
traceGroup@durationInNanos
```
that way we do not lose the compatibility. Alternatively, we use a different prefix e.g.
```
traceGroupInfo.name
traceGroupInfo.statusCode
traceGroupInfo.endTime
traceGroupInfo.durationInNanos
```
But then one has to hardcode the population of both `traceGroup` and `traceGroupInfo.name` to support backward compatibility, which I do not think is a good design. Can discuss further offline.

chenqi0805: BTW, I tested that data-ingestion works fine with the sample app (given that no old data is in the ODFE). But trace-analytics-ui breaks as expected.

kowshikn: Yeah. Can we actually keep `traceGroup` as key word and add new property called `traceGroupFields`. Also, you should check with @joshuali925 .

chenqi0805: Make the change and the compatibility issue is resolved by combining together with https://github.com/opendistro-for-elasticsearch/trace-analytics/pull/33

wrijeff (approved):
👍

Issue #476, opened 2021-04-02
Ensure Shutdownhook works with Peerforwarding


kowshikn:
On shutdown/startup of data prepper instance in a multi-node setup, should shutdown gracefully by processing all the cached data in a stateful plugin.

PR #512, opened 2021-04-13
Revised shutdown behavior to limit data loss.

wrijeff:
*Issue #, if available:* #476

*Description of changes:*
Need to ensure pipelines and components are shut down in a graceful manner to prevent data loss (specifically within stateful components). Added 2 new APIs to the Prepper interface: `prepareForShutdown()` and `isReadyForShutdown()`. High level flow:
1. Shutdown API is called on the server
2. `shutdown()` is called on each Pipeline. Pipelines are sorted in a topological order such that shutdown occurs starting upstream and proceeding downstream.
3. Within each pipeline:
    1. The source is stopped to prevent new data from entering
    2. Preppers are notified of the shutdown via `prepareForShutdown()`. This allows them time to clear state by flushing records
        * There's still an improvement here to be made when dealing with long chains of preppers, but it can wait
    3. The pipeline waits for ProcessWorkers to exit their work loop. This happens once the buffer is clear and every prepper returns true when `isReadyForShutdown()` is called.
    4. ProcessWorkers are stopped
    5. Preppers and sinks are `shutdown()` (cleans up any resources)
    6. A new sinkExecutorService is stopped. 
        * Previously a single prepperSinkExecutorService was used for both preppers and sinks, however it had to be split because of a circular dependency. `prepperSinkExecutorService` was used to run ProcessWorkers, but ProcessWorkers also used it to submit sink-write jobs. So "shutting prepperSinkExecutorService down" would wait for ProcessWorkers to exit their loops, but part of exiting their loops included submitting new tasks to the prepperSinkExecutorService when writing to sinks.

Implemented the shutdown flow described above, as well as implemented the new methods and tests in existing preppers. Also misc fixes in the ServiceMap prepper tests.

Testing calling `/shutdown` on the sample app with 4 workers per pipeline:
```
2021-04-13T18:15:50,266 [entry-pipeline-process-worker-1-thread-4] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  entry-pipeline Worker: Processing 1 records from buffer
2021-04-13T18:15:50,367 [service-map-pipeline-process-worker-3-thread-3] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  service-map-pipeline Worker: Processing 1 records from buffer
2021-04-13T18:15:51,866 [raw-pipeline-process-worker-5-thread-3] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  raw-pipeline Worker: Processing 2 records from buffer
2021-04-13T18:15:51,871 [raw-pipeline-process-worker-5-thread-4] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  raw-pipeline Worker: Processing 2 records from buffer
2021-04-13T18:15:51,873 [raw-pipeline-process-worker-5-thread-1] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  raw-pipeline Worker: Processing 3 records from buffer
2021-04-13T18:15:53,267 [raw-pipeline-process-worker-5-thread-2] INFO  com.amazon.dataprepper.pipeline.ProcessWorker -  raw-pipeline Worker: Processing 3 records from buffer
2021-04-13T18:15:57,500 [HTTP-Dispatcher] INFO        com.amazon.dataprepper.DataPrepper - Shutting down pipeline: entry-pipeline
2021-04-13T18:15:57,501 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [entry-pipeline] - Received shutdown signal with timeout 10000, will initiate the shutdown process
2021-04-13T18:15:59,523 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.plugins.source.oteltrace.OTelTraceSource - Stopped otel_trace_source.
2021-04-13T18:15:59,524 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [entry-pipeline] - Shutting down process workers
2021-04-13T18:15:59,582 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [entry-pipeline] - Shutting down process workers
2021-04-13T18:15:59,583 [HTTP-Dispatcher] INFO        com.amazon.dataprepper.DataPrepper - Shutting down pipeline: service-map-pipeline
2021-04-13T18:15:59,583 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [service-map-pipeline] - Received shutdown signal with timeout 10000, will initiate the shutdown process
2021-04-13T18:15:59,583 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [service-map-pipeline] - Shutting down process workers
2021-04-13T18:15:59,775 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [service-map-pipeline] - Shutting down process workers
2021-04-13T18:15:59,775 [HTTP-Dispatcher] INFO        com.amazon.dataprepper.DataPrepper - Shutting down pipeline: raw-pipeline
2021-04-13T18:15:59,775 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [raw-pipeline] - Received shutdown signal with timeout 10000, will initiate the shutdown process
2021-04-13T18:15:59,775 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.plugins.prepper.oteltrace.OTelTraceRawPrepper - Preparing for shutdown, re-enqueueing 12 spans
2021-04-13T18:15:59,780 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [raw-pipeline] - Shutting down process workers
2021-04-13T18:16:02,270 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.Pipeline - Pipeline [raw-pipeline] - Shutting down process workers
2021-04-13T18:16:02,271 [HTTP-Dispatcher] INFO  com.amazon.dataprepper.pipeline.server.DataPrepperServer - Data Prepper server stopped

```

-----

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
@@ -21,8 +21,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -54,7 +54,9 @@ public class PipelineParser {
                     new TypeReference<Map<String, PipelineConfiguration>>() {
                     });
             final List<String> allPipelineNames = PipelineConfigurationValidator.validateAndGetPipelineNames(pipelineConfigurationMap);
-            final Map<String, Pipeline> pipelineMap = new HashMap<>();
+
+            // LinkedHashMap to preserve insertion order
+            final Map<String, Pipeline> pipelineMap = new LinkedHashMap<>();
             pipelineConfigurationMap.forEach((pipelineName, configuration) ->
                     configuration.updateCommonPipelineConfiguration(pipelineName));
             for (String pipelineName : allPipelineNames) {
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
@@ -7,7 +7,6 @@ import com.amazon.dataprepper.model.sink.Sink;
 import com.amazon.dataprepper.model.source.Source;
 import com.amazon.dataprepper.pipeline.common.PipelineThreadFactory;
 import com.amazon.dataprepper.pipeline.common.PipelineThreadPoolExecutor;
-import com.amazon.dataprepper.plugins.buffer.blockingbuffer.BlockingBuffer;
 import com.google.common.base.Preconditions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -31,8 +30,8 @@ import static java.lang.String.format;
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class Pipeline {
     private static final Logger LOG = LoggerFactory.getLogger(Pipeline.class);
-    private static final int PREPPER_DEFAULT_TERMINATION_IN_MILLISECONDS = 5000;
-    private boolean stopRequested;
+    private static final int PREPPER_DEFAULT_TERMINATION_IN_MILLISECONDS = 10_000;
+    private volatile boolean stopRequested;
 
     private final String name;
     private final Source source;
@@ -41,7 +40,8 @@ public class Pipeline {
     private final List<Sink> sinks;
     private final int prepperThreads;
     private final int readBatchTimeoutInMillis;
-    private final ExecutorService prepperSinkExecutorService;
+    private final ExecutorService prepperExecutorService;
+    private final ExecutorService sinkExecutorService;
 
     /**
      * Constructs a {@link Pipeline} object with provided {@link Source}, {@link #name}, {@link Collection} of
@@ -76,12 +76,14 @@ public class Pipeline {
         this.sinks = sinks;
         this.prepperThreads = prepperThreads;
         this.readBatchTimeoutInMillis = readBatchTimeoutInMillis;
-        int coreThreads = sinks.size() + prepperThreads; //TODO We may have to update this after benchmark tests
-        this.prepperSinkExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(coreThreads,
+        this.prepperExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(prepperThreads,
                 new PipelineThreadFactory(format("%s-process-worker", name)), this);
+
+        this.sinkExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(sinks.size(),
+                new PipelineThreadFactory(format("%s-process-worker", name)), this);
+
         stopRequested = false;
     }
-
 
     /**
      * @return Unique name of this pipeline.
@@ -146,7 +148,7 @@ public class Pipeline {
                             }
                         }
                 ).collect(Collectors.toList());
-                prepperSinkExecutorService.submit(new ProcessWorker(buffer, preppers, sinks, this));
+                prepperExecutorService.submit(new ProcessWorker(buffer, preppers, sinks, this));
             }
         } catch (Exception ex) {
             //source failed to start - Cannot proceed further with the current pipeline, skipping further execution
@@ -162,7 +164,13 @@ public class Pipeline {
     }
 
     /**
-     * Initiates shutdown of the pipeline by notifying the components to stop processing.
+     * Initiates shutdown of the pipeline by:
+     * 1. Stopping the source to prevent new items from being consumed
+     * 2. Notifying preppers to prepare for shutdown (e.g. flushing batched items)
+     * 3. Waiting for ProcessWorkers to exit their run loop (only after buffer/preppers are empty)
+     * 4. Stopping the ProcessWorkers if they are unable to exit gracefully
+     * 5. Shutting down preppers and sinks
+     * 6. Stopping the sink ExecutorService
      *
      * @param prepperTimeout the maximum time to wait after initiating shutdown to forcefully shutdown process worker
      */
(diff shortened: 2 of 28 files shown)


data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
```diff
@@ -54,7 +54,9 @@ public PipelineParser(final String configurationFileLocation) {
                     new TypeReference<Map<String, PipelineConfiguration>>() {
                     });
             final List<String> allPipelineNames = PipelineConfigurationValidator.validateAndGetPipelineNames(pipelineConfigurationMap);
-            final Map<String, Pipeline> pipelineMap = new HashMap<>();
+
+            // LinkedHashMap to preserve insertion order
+            final Map<String, Pipeline> pipelineMap = new LinkedHashMap<>();
```
chenqi0805, line 59: QUES: Is the insertion order guaranteed to be topological? It seems a little fragile but have not checked the context carefully.

wrijeff: It is, but only because we're calling [validateAndGetPipelineNames](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/e2b858c0f15928bf28857d44d8b4f567aa973f49/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineConfigurationValidator.java#L22-L31) which does the sorting for us.

I agree it's a bit fragile - maybe a better approach would have the `shutdown` logic perform the sorting just in case the "validate" method ever stops sorting. I used what already existed here though as a shortcut.

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
```diff
@@ -247,6 +249,37 @@ private boolean shouldGarbageCollect() {
...
+     */
+    @Override
+    public void prepareForShutdown() {
+        boolean isLockAcquired = prepareForShutdownLock.tryLock();
+
+        if (isLockAcquired) {
+            try {
+                LOG.info("Preparing for shutdown, re-enqueueing {} spans", delayedParentSpanQueue.size());
+                Iterator iterator = delayedParentSpanQueue.iterator();
+                List<DelayedParentSpan> delayedParentSpanList = ImmutableList.copyOf(iterator);
```
chenqi0805, line 263: QUES: is there reason we do not use drainTo here?

chenqi0805 (commented):
One high level comment: Is it better we add default implementation of prepareforShutdown and isReadyforShutdown in AbstractPrepper? I can see repeated trivial implementation in the specific plugins:

```
@Override
        public void prepareForShutdown() {

        }

        @Override
        public boolean isReadyForShutdown() {
            return true;
        }
```

data-prepper-plugins/otel-trace-raw-prepper/src/main/java/com/amazon/dataprepper/plugins/prepper/oteltrace/OTelTraceRawPrepper.java
wrijeff: Yes - drainTo only transfers objects that meet the delay criteria. 

Using an iterator allows us to fetch ALL objects in the queue regardless of the delay values.

wrijeff:
> One high level comment: Is it better we add default implementation of prepareforShutdown and isReadyforShutdown in AbstractPrepper? I can see repeated trivial implementation in the specific plugins:
> 
> ```
> @Override
>         public void prepareForShutdown() {
> 
>         }
> 
>         @Override
>         public boolean isReadyForShutdown() {
>             return true;
>         }
> ```

Personally I prefer this way to force new implementers to explicitly acknowledge that "no action is required" when `prepareForShutdown` is called. 

It just so happens that currently many preppers are either simple (no-op, String, etc.) or stateless, meaning that their shutdown behavior is trivial.

Maybe a better abstraction would be to only require a StatefulPrepper class to implement these methods? Regardless, I think the extra bit of duplication is worth the explicitness.

chenqi0805 (approved):

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
yadavcbala: I do not see a use case where we will disable the validate, why are we calling it fragile ? I agree that when we implemented it, we implemented as a bonus but not as a requirement. 
Although it ensures that pipelines are always in topological order, there is a case where if we have orphan pipelines - each pipeline will be in topological order but pipeline ids can be interleaved, which I think will not be an issue - if we are shutting down everything.

Example: [_very rare scenario but possible depending on the order of pipelines defined in yml file_]
P1 -> P2 -> P3 -> P4
Pa -> Pb -> Pc
The order can be like  -> P1, P2, Pa, P3, Pb, P4, Pc.

We also had a commented method to validate orphans and fail but I think we removed it.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
```diff
@@ -76,13 +76,15 @@ public Pipeline(
         this.sinks = sinks;
         this.prepperThreads = prepperThreads;
         this.readBatchTimeoutInMillis = readBatchTimeoutInMillis;
-        int coreThreads = sinks.size() + prepperThreads; //TODO We may have to update this after benchmark tests
-        this.prepperSinkExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(coreThreads,
+        this.prepperExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(prepperThreads,
                 new PipelineThreadFactory(format("%s-process-worker", name)), this);
+
+        this.sinkExecutorService = PipelineThreadPoolExecutor.newFixedThreadPool(sinks.size(),
+                new PipelineThreadFactory(format("%s-process-worker", name)), this);
```
yadavcbala, line 83: assuming it is intentional to name sink threads also as `...process-worker...`

yadavcbala (approved):
LGTM - one question

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/Pipeline.java
wrijeff: Was not intentional, good point. Will changes names to be more indicative of preppers/sinks.

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/PipelineParser.java
wrijeff: Ah the logic is fine - I'm glad `validate` was implemented, it saved me a lot of effort :) Also interleaved is fine too, just as long as each individual pipeline chain is shutdown in-order from upstream to downstream.

We're calling my implementation here "fragile" because currently there exists a slight temporal coupling between `validateAndGetPipelineNames` and `shutdown`. In this case, `shutdown` _relies_ on some other operation to perform the topological sort and persist those values in the hashmap. `shutdown` has a soft requirement that `validate` be called prior. It would be more robust if `shutdown` instead performed whichever actions it needed (sorting) as part of its routine.

I agree it's unlikely, but if for whatever reason someone comes along and does a refactor which causes the pipelineMap to no longer be sorted, `shutdown`'s behavior will be impacted in a non-obvious way.

PR #540, opened 2021-04-21
String attributes for boolean, long, int for pipeline yaml config

nclaveeoh-amzn:
This change is to allow the yaml property values for boolean, integer, and long values to be passed in as strings. This is supported for elements in the data-prepper config and some of the elements in the pipeline config. Nulls values are also now supported. Custom elements in source and sinks do not support this. It is not supported because the parsing of those configurations is done with custom code while the data-prepper config supports it because it uses json property binding via Jackson libraries. This changes brings more consistency throughout the config files.

String values were previously not supported but now they are:
```
entry-pipeline: 
  source: 
    otel_trace_source:  
      health_check_service: "false"  
      port: "8080"  
```

Null values were previously not supported but now they are:
```
entry-pipeline: 
  source: 
    otel_trace_source:  
      health_check_service:
      port:
```

Continues to be supported:
```
entry-pipeline: 
  source: 
    otel_trace_source:  
      health_check_service: false
      port: 8080
```

-----

### DIFF
--- a/data-prepper-api/src/main/java/com/amazon/dataprepper/model/configuration/PluginSetting.java
+++ b/data-prepper-api/src/main/java/com/amazon/dataprepper/model/configuration/PluginSetting.java
@@ -4,6 +4,8 @@ import java.util.Map;
 
 public class PluginSetting {
 
+    private static final String UNEXPECTED_ATTRIBUTE_TYPE_MSG = "Unexpected type [%s] for attribute [%s]";
+
     private final String name;
     private final Map<String, Object> settings;
     private int processWorkers;
@@ -100,7 +102,7 @@ public class PluginSetting {
             return Integer.valueOf(String.valueOf(object));
         }
 
-        throw new IllegalArgumentException("Unexpected type for attribute: " + attribute);
+        throw new IllegalArgumentException(String.format(UNEXPECTED_ATTRIBUTE_TYPE_MSG, object.getClass(), attribute));
     }
 
     /**
@@ -120,7 +122,7 @@ public class PluginSetting {
             return String.valueOf(object);
         }
 
-        throw new IllegalArgumentException("Unexpected type for attribute: " + attribute);
+        throw new IllegalArgumentException(String.format(UNEXPECTED_ATTRIBUTE_TYPE_MSG, object.getClass(), attribute));
     }
 
     /**
@@ -142,7 +144,7 @@ public class PluginSetting {
             return Boolean.valueOf(String.valueOf(object));
         }
 
-        throw new IllegalArgumentException("Unexpected type for attribute: " + attribute);
+        throw new IllegalArgumentException(String.format(UNEXPECTED_ATTRIBUTE_TYPE_MSG, object.getClass(), attribute));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


wrijeff:
@nclaveeoh-amzn: We're starting to require commit signing (it's in the [contributing doc](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) but buried), I'll see how we can make it more visible. 

For now can you amend the commit with the signature? It should be something like:
1. `git commit --amend -s`
2. `git push -f origin <branch_name>`

data-prepper-api/src/test/java/com/amazon/dataprepper/model/configuration/PluginSettingsTests.java
```diff
@@ -3,79 +3,160 @@
...
+
+        assertThat(pluginSetting.getAttributeFromSettings(TEST_INT_ATTRIBUTE), is(TEST_INT_VALUE));
+
+        // test attributes that exist when passing in a different default value
+        assertThat(pluginSetting.getAttributeOrDefault(TEST_INT_ATTRIBUTE, TEST_INT_DEFAULT_VALUE), is(TEST_INT_VALUE));
+        assertThat(pluginSetting.getIntegerOrDefault(TEST_INT_ATTRIBUTE, TEST_INT_DEFAULT_VALUE), is(TEST_INT_VALUE));
+        assertThat(pluginSetting.getStringOrDefault(TEST_STRING_ATTRIBUTE, TEST_STRING_DEFAULT_VALUE),
+                is(equalTo(TEST_STRING_VALUE)));
+        assertThat(pluginSetting.getBooleanOrDefault(TEST_BOOL_ATTRIBUTE, TEST_BOOL_DEFAULT_VALUE), is(equalTo(TEST_BOOL_VALUE)));
+        assertThat(pluginSetting.getLongOrDefault(TEST_LONG_ATTRIBUTE, TEST_LONG_DEFAULT_VALUE), is(equalTo(TEST_LONG_VALUE)));
```
wrijeff, line 72: nit - Personal preference would be to split these larger tests up into multiple tests per type (string/long/bool/int), but seeing as we don't expect this class to change much it works

data-prepper-api/src/main/java/com/amazon/dataprepper/model/configuration/PluginSetting.java
```diff
@@ -88,49 +88,83 @@ public Object getAttributeOrDefault(final String attribute, final Object default
...
+        Object object = getAttributeOrDefault(attribute, defaultValue);
+        if (object == null) {
+            return null;
+        } else if (object instanceof Number) {
+            return ((Number) object).intValue();
+        } else if (object instanceof String) {
+            return Integer.valueOf(String.valueOf(object));
+        }
+
+        throw new IllegalArgumentException("Unexpected type for attribute: " + attribute);
```
wrijeff, line 103: Can the invalid class be added to the error message with `attribute.getClass()`?

wrijeff (approved):
👍

kowshikn (approved):

chenqi0805 (approved):
LGTM! Thanks for the contribution

data-prepper-api/src/test/java/com/amazon/dataprepper/model/configuration/PluginSettingsTests.java
nclaveeoh-amzn: i agree with that structure, will do that

Issue #574, opened 2021-05-06
Apply automatic code-formatting to the repo


wrijeff:
Check out applying the [spotless gradle plugin](https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md) to the repo, potentially integrate it to the build process.

PR #584, opened 2021-05-11
MAINT: add spotless code formatting -- first pass

chenqi0805:
*Issue #, if available:*
#574 

*Description of changes:*

-----
This PR 

(1) Applied spotless code formatting to all file types
(2) Introduced the basic rules.


chenqi0805:
Get a blocking bug:
```
Cannot cast object 'extension 'docker'' with class 'com.bmuschko.gradle.docker.DockerExtension_Decorated' to class 'com.bmuschko.gradle.docker.DockerExtension'
```
It seems to stem from some conflict between spotless and existing com.bmuschko.docker-remote-api plugin.

wrijeff:
> Get a blocking bug:
> 
> ```
> Cannot cast object 'extension 'docker'' with class 'com.bmuschko.gradle.docker.DockerExtension_Decorated' to class 'com.bmuschko.gradle.docker.DockerExtension'
> ```
> 
> It seems to stem from some conflict between spotless and existing com.bmuschko.docker-remote-api plugin.

Appears to have the same root cause as https://github.com/opendistro-for-elasticsearch/data-prepper/pull/551. Not sure how the upgraded version is getting pulled in on this PR though. Fixing the prior issue should unblock this one.

chenqi0805:
Turns out by removing 
```
plugins {
    id 'com.bmuschko.docker-remote-api' version '6.6.1'
}
```
from the root gradle file `data-prepper-core/build.gradle` the issus is resolved.

build.gradle
```diff
@@ -1,17 +1,38 @@
...
     repositories {
         mavenCentral()
         maven { url "https://jitpack.io" }
     }
+
+    spotless {
+        format 'misc', {
+            target '*.md', '.gitignore', '*.yml', '*.yaml'
+            // TODO: enrich format rules
+            trimTrailingWhitespace()
```
wrijeff, line 20: We might want to _not_ trim whitespace on markdown files. From https://daringfireball.net/projects/markdown/syntax#block:

>"When you do want to insert a \<br /> break tag using Markdown, you end a line with two or more spaces, then type return."

Would just need to confirm if the formatter would strip double-spaces which were intended to be used as line breaks. Or we can just encourage people to not use double-spaces in our files. 😉 

I only see one double-space being removed in this PR, and I don't think it was intended to be a line break. Can probably merge as is and test locally.

wrijeff (approved):
👍

build.gradle
chenqi0805: > We might want to not trim whitespace on markdown files. From https://daringfireball.net/projects/markdown/syntax#block:

Sure. Will reserve markdown in a separate block.

> Would just need to confirm if the formatter would strip double-spaces which were intended to be used as line breaks. Or we can just encourage people to not use double-spaces in our files.

No. Tested it does not.

Issue #575, opened 2021-05-07
[Metrics] Make metrics MeterRegistry modular and configurable 


yadavcbala:
**Is your feature request related to a problem? Please describe.**
No, It is not related to a problem but allows users to configure MeterRegistry using a property in server configuration file.

**Describe the solution you'd like**
At the time of this writing, data prepper has default MeterRegistry as Prometheus i.e. metrics generated by data prepper are accessible via data prepper server at `/metrics/prometheus` but there might be users with need to publish these metrics to CloudWatch. As part of #566 we have added CloudWatchMeterRegistry and configuring it would require updating code in multiple places (_see below for changes required_), this issue is to enable configuring MeterRegistry as part of server configuration (refer Server Configuration section [here](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/main/docs/readme/configuration.md))

**Additional context**
Changes required for configuring CloudWatchMeterRegistry in place of PrometheusMeterRegistry

- Update DataPrepper.java 
```
//update sysJVMMeterRegistry
private static final CloudWatchMeterRegistry sysJVMMeterRegistry = new CloudWatchMeterRegistryProvider().getCloudWatchMeterRegistry();

public static CloudWatchMeterRegistry getSysJVMMeterRegistry() {
        return sysJVMMeterRegistry;
    }

//the existing method startPrometheusBackend() needs to be replaced with
    private static void startCloudWatchMeterRegistry() {
        final CloudWatchMeterRegistry prometheusMeterRegistry = new CloudWatchMeterRegistryProvider().getCloudWatchMeterRegistry();
        Metrics.addRegistry(prometheusMeterRegistry);
    }
```

- Update
```
//Remove
server.createContext("/metrics/prometheus", new PrometheusMetricsHandler());
server.createContext("/metrics/sys", new PrometheusMetricsHandler(DataPrepper.getSysJVMMeterRegistry()));
```

PR #585, opened 2021-05-12
Add capability to configure MeterRegistry via server configuration file

yadavcbala:
*Issue #, if available:*
#575 

*Description of changes:*
* Add capability to configure MeterRegistry from server configuration file
* Add capability to configure multiple meter registries
* Update documentation accordingly

*Testing*
* Verified using default PrometheusMeterRegistry  
* Verified using both Prometheus and CloudWatch
-----


data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
```diff
@@ -42,12 +48,14 @@ public DataPrepperConfiguration(
             @JsonProperty("keyStoreFilePath") final String keyStoreFilePath,
             @JsonProperty("keyStorePassword") final String keyStorePassword,
             @JsonProperty("privateKeyPassword") final String privateKeyPassword,
-            @JsonProperty("serverPort") final String serverPort
+            @JsonProperty("serverPort") final String serverPort,
+            @JsonProperty("metricsRegistry") final List<MeterRegistryType> metricsRegistry
```
chenqi0805, line 52: Nit: plural in both user exposed and variable names

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
```diff
@@ -71,6 +79,15 @@ public String getPrivateKeyPassword() {
         return privateKeyPassword;
     }
 
+    public List<MeterRegistryType> getMetricsRegistry() {
```
chenqi0805, line 82: Nit: plural

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -40,12 +48,24 @@ public DataPrepperServer(final DataPrepper dataPrepper) {
             throw new RuntimeException("Failed to create server", e);
         }
 
-        server.createContext("/metrics/prometheus", new PrometheusMetricsHandler());
-        server.createContext("/metrics/sys", new PrometheusMetricsHandler(DataPrepper.getSysJVMMeterRegistry()));
+        if (dataPrepperConfiguration.isMeterRegistryTypeConfigured(MeterRegistryType.Prometheus)) {
+            final PrometheusMeterRegistry prometheusMeterRegistryForDataPrepper =
+                    getPrometheusMeterRegistryFromRegistries(Metrics.globalRegistry.getRegistries());
+            server.createContext("/metrics/prometheus", new PrometheusMetricsHandler(prometheusMeterRegistryForDataPrepper));
```
chenqi0805, line 54: Ques: is it possible to scrape from `CloudWatchMeterRegistry`? Is there any outstanding difference? It seems more consistent to me to use the user configured meter registry type for scraping. But that can also happen later.

yadavcbala: CloudWatchMeterRegistry publishes to CloudWatch - no scrapping possible

chenqi0805 (approved):
LGTM

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
```diff
@@ -71,6 +78,15 @@ public String getPrivateKeyPassword() {
         return privateKeyPassword;
     }
 
+    public List<MeterRegistryType> getMetricsRegistries() {
+        return metricsRegistries;
```
wrijeff, line 82: Naming - `metricsRegistries` to `meterRegistries`? Applies to variable definitions at top of class too.

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
```diff
@@ -71,6 +78,15 @@ public String getPrivateKeyPassword() {
...
+    }
+
+    public boolean isMeterRegistryTypeConfigured(final MeterRegistryType meterRegistryType) {
```
wrijeff, line 85: I see this method is used a bit further down to check if a Prometheus registry exists, then `getPrometheusMeterRegistryFromRegistries` is called to do a similar search (but return the object instead of a boolean).

Could this method be changed to something like `Optional<MeterRegistry> getRegisteredMeterByType(...)`? Callers could then `result.isPresent()` to get the same bool functionality.

data-prepper-core/src/main/java/com/amazon/dataprepper/pipeline/server/DataPrepperServer.java
```diff
@@ -40,12 +48,24 @@ public DataPrepperServer(final DataPrepper dataPrepper) {
             throw new RuntimeException("Failed to create server", e);
         }
 
-        server.createContext("/metrics/prometheus", new PrometheusMetricsHandler());
-        server.createContext("/metrics/sys", new PrometheusMetricsHandler(DataPrepper.getSysJVMMeterRegistry()));
+        if (dataPrepperConfiguration.isMeterRegistryTypeConfigured(MeterRegistryType.Prometheus)) {
```
wrijeff, line 51: Feels kind of off to have an explicit check here, but should be fine unless the types of meters and required server actions grows. Probably too much effort for little gain.

wrijeff (approved):
👍 - just left a few style questions/comments

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
```diff
@@ -27,32 +32,23 @@
...
 
     /**
-     * Set the DataPrepperConfiguration from a file
+     * Set the DataPrepperConfiguration from file, if exists; else use default configuration
      * @param configurationFile File containing DataPrepperConfiguration yaml
      */
-    public static void configure(final String configurationFile) {
-        final DataPrepperConfiguration dataPrepperConfiguration =
+    public static void configure(@Nullable final String configurationFile) {
+        configuration = configurationFile == null ? DataPrepperConfiguration.DEFAULT_CONFIG :
```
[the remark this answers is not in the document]
yadavcbala: I thought about the Optional<String> but then the suggestion is not to use Optional as parameters/arguments. I will attempt to add parameter-less `configureWithDefaults()`

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/MeterRegistryType.java
```diff
@@ -0,0 +1,24 @@
...
+        MeterRegistry meterRegistry = null;
+        switch (meterRegistryType) {
+            case Prometheus:
+                meterRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
+                break;
+            case CloudWatch:
+                meterRegistry = new CloudWatchMeterRegistryProvider().getCloudWatchMeterRegistry();
+                break;
+        }
+        return meterRegistry;
```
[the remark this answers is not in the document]
yadavcbala: I was not inclined for default case thinking we may want to proceed without Metrics, but may be not - thinking again, it sounds like either default or configured, if we see issues in configured, then exception

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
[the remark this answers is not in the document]
wrijeff: Good point - I shouldn't have suggested an Optional as a parameter. It's better to avoid method behavior splitting based on null/non-null args if possible.

data-prepper-core/src/main/java/com/amazon/dataprepper/DataPrepper.java
```diff
@@ -69,20 +74,44 @@ private DataPrepper() {
...
+        switch (metricRegistryType) {
+            case Prometheus:
+                Metrics.addRegistry(new PrometheusMeterRegistry(PrometheusConfig.DEFAULT));
+                break;
+            case CloudWatch:
+                Metrics.addRegistry(new CloudWatchMeterRegistryProvider().getCloudWatchMeterRegistry());
+                break;
+            default:
+                throw new IllegalArgumentException(format("Invalid metricRegistryType %s", metricRegistryType));
+        }
```
wrijeff, line 100: Just now noticed this - could we remove this switch block in favor of `MetricRegistryType::getDefaultMeterRegistryForType`?

Issue #491, opened 2021-04-08
invalid_alias_name_exception on Data Prepper startup


wrijeff:
**Describe the bug**
An ES cluster got into a weird state when an index was created named `otel-v1-apm-span` even though this name is intended to be reserved for an index _alias_. 

```
{"error":{"root_cause":[{"type":"invalid_alias_name_exception","reason":"Invalid alias name [otel-v1-apm-span], an index exists with the same name as the alias","index_uuid":"0-ZhKKdyRHeJjw4QKjF-Yg","index":"otel-v1-apm-span"}],"type":"invalid_alias_name_exception","reason":"Invalid alias name [otel-v1-apm-span], an index exists with the same name as the alias","index_uuid":"0-ZhKKdyRHeJjw4QKjF-Yg","index":"otel-v1-apm-span"},"status":400}
```
Refer to #479 for more information. We should determine how to get to this invalid state and find if there's anything we can do to code around it. Or at the very least provide a more useful error message indicating that the index needs to be manually deleted.

chenqi0805:
steps to reproduce the bug by `examples/dev/trace-analytics-sample-app`:

1. start all containers as normal: `docker-compose up -d`
2. In the Kibana interactive console, run `DELETE otel-v1-apm-span-*`
3. open `localhost:8089` and start sending traces by clicking the buttons. verify in the kibana dashboard that an index was created named `otel-v1-apm-span`: 
![Screen Shot 2021-05-12 at 2 28 46 PM](https://user-images.githubusercontent.com/19492223/118034373-85873e80-b32f-11eb-8c99-58175cb30d82.png)

4. docker restart data-prepper container
5. check data-prepper log: 
```
java.lang.reflect.InvocationTargetException: null
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:64) ~[?:?]
        at jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:?]
        at java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500) ~[?:?]
        at java.lang.reflect.Constructor.newInstance(Constructor.java:481) ~[?:?]
[...]

PR #588, opened 2021-05-14
Bug Fixes: capture invalid index alias exception and display instructive msg

chenqi0805:
Signed-off-by: qchea <qchea@amazon.com>

*Issue #, if available:*
#491 

*Description of changes:*

-----
This PR:

(1) Captures invalid raw-span index alias caused by existing index under the same name from misoperation on ES backend.
(2) make sure REST client gets cleaned up if ElasticsearchSink starts failed to prevent thread leakage.

### DIFF
--- a/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
+++ b/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
@@ -71,10 +71,9 @@ public class ElasticsearchSink extends AbstractSink<Record<String>> {
     this.documentIdField = esSinkConfig.getIndexConfiguration().getDocumentIdField();
     try {
       start();
-    } catch (final IOException e) {
-      throw new RuntimeException(e.getMessage(), e);
-    } finally {
+    } catch (final Exception e) {
       this.shutdown();
+      throw new RuntimeException(e.getMessage(), e);
     }
   }
 
--- a/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
+++ b/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
@@ -46,6 +46,7 @@ public class ElasticsearchSink extends AbstractSink<Record<String>> {
   private static final Logger LOG = LoggerFactory.getLogger(ElasticsearchSink.class);
   // Pulled from BulkRequest to make estimation of bytes consistent
   private static final int REQUEST_OVERHEAD = 50;
+  protected static final String INDEX_ALIAS_USED_AS_INDEX_ERROR = "Invalid alias name [%s], an index exists with the same name as the alias";
 
   private BufferedWriter dlqWriter;
   private final ElasticsearchSinkConfiguration esSinkConfig;
@@ -72,6 +73,8 @@ public class ElasticsearchSink extends AbstractSink<Record<String>> {
       start();
     } catch (final IOException e) {
       throw new RuntimeException(e.getMessage(), e);
+    } finally {
+      this.shutdown();
     }
   }
 
@@ -187,6 +190,10 @@ public class ElasticsearchSink extends AbstractSink<Record<String>> {
         if (e.getMessage().contains("resource_already_exists_exception")) {
           // Do nothing - likely caused by a race condition where the resource was created
           // by another host before this host's restClient made its request
+        } else if (e.getMessage().contains(String.format(INDEX_ALIAS_USED_AS_INDEX_ERROR, indexAlias))) {
+          throw new ElasticsearchException(
+                  String.format("An index exists with the same name as the reserved index alias name [%s], please delete or migrate the existing index",
+                          indexAlias));
         } else {
           throw e;
         }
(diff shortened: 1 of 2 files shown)


data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -70,8 +71,12 @@ public ElasticsearchSink(final PluginSetting pluginSetting) {
...
     try {
       start();
-    } catch (final IOException e) {
-      throw new RuntimeException(e.getMessage(), e);
+    } catch (final Exception e) {
+      try {
+        throw new RuntimeException(e.getMessage(), e);
+      } finally {
+        this.shutdown();
+      }
```
wrijeff, line 79: Why change the exception type here?

Also can't we just add a `finally` block to the original try/catch, instead of introducing a new try/finally?

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -187,6 +192,10 @@ private void checkAndCreateIndex() throws IOException {
         if (e.getMessage().contains("resource_already_exists_exception")) {
           // Do nothing - likely caused by a race condition where the resource was created
           // by another host before this host's restClient made its request
+        } else if (e.getMessage().contains(String.format(INDEX_ALIAS_USED_AS_INDEX_ERROR, indexAlias))) {
+          throw new ElasticsearchException(
+                  String.format("An index exists with the same name as the reserved index alias name [%s], please delete or migrate the existing index",
```
wrijeff, line 197: We should consider defining custom Data Prepper exception types instead of re-using ES's (can be another maintenance task).

I'd imagine we would define base DataPrepperException's in the API package, and have plugins define extensions where needed.

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
chenqi0805: > Why change the exception type here?

If you mean from IOException to Exception, this is to capture `ElasticsearchException` from `start()`.

> Also can't we just add a finally block to the original try/catch, instead of introducing a new try/finally?
The main reason is that finally will close the REST Client even if no exception is detected. Thus I have to do this. example:

```
try {
    System.out.println("a");
} finally {
    System.out.println("b");
}
```
both a and b will be printed.

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
chenqi0805: Great call. We could wrap it into a checked data-prepper exception. For now I think I will wrap it into IOException.

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
wrijeff: Ah, I misinterpreted `start()` to be a long-running operation, something to be interrupted. This is more like a initialization method. Potentially suggest renaming the method as such (doesn't have to be now).

How about just 
```
this.shutdown();
throw new RuntimeException(e.getMessage(), e);
```
to avoid the new try/finally then?

wrijeff (approved):
👍

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
chenqi0805: Will do renaming.

> How about just
```
this.shutdown();
throw new RuntimeException(e.getMessage(), e);
```
to avoid the new try/finally then?

This is what I initially do. Then I realized any runtime exception from `shutdown` could overwrite the captured exception. But I probably overthought here.

data-prepper-plugins/elasticsearch/src/test/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSinkIT.java
```diff
@@ -113,6 +113,15 @@ public void testInstantiateSinkRawSpanDefault() throws IOException {
     }
   }
 
+  public void testInstantiateSinkRawSpanReservedAliasAlreadyUsedAsIndex() throws IOException {
+    final String reservedIndexAlias = IndexConstants.TYPE_TO_DEFAULT_ALIAS.get(IndexConstants.RAW);
+    final Request request = new Request(HttpMethod.PUT, reservedIndexAlias);
+    client().performRequest(request);
+    final PluginSetting pluginSetting = generatePluginSetting(true, false, null, null);
+    assertThrows(String.format(ElasticsearchSink.INDEX_ALIAS_USED_AS_INDEX_ERROR, reservedIndexAlias),
+            RuntimeException.class, () -> new ElasticsearchSink(pluginSetting));
```
wrijeff, line 122: Assert IOException here?

wrijeff (approved):
👍

Issue #569, opened 2021-05-05
Add schema version to index templates


wrijeff:
Data Prepper should check the version of index templates on startup before making a POST call. Templates have a "version" field, but it is limited to just an integer.

The new `_index_template` ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/index-templates.html)) allows for metadata, see if we could switch over from legacy templates. This would allow us to specify our own customer metadata, like `schema_version: 2.12`.

1. Determine if using `_index_template` would provide same functionality we're currently getting out of `_template`
2. Determine if the Java client supports making `_index_template` requests (appears to be via `PutComposableIndexTemplateRequest`, see https://www.elastic.co/guide/en/elasticsearch/client/java-rest/7.10/java-rest-high-put-index-template-v2.html)
3. Determine amount of effort required to switch to new template style (also would likely require front-end changes)

PR #608, opened 2021-05-17
Added version to index template schemas.

wrijeff:
Signed-off-by: Jeff Wright <74204404+wrijeff@users.noreply.github.com>

*Issue #, if available:* #569

*Description of changes:*
* Added logic in ES Sink to check version of index templates before attempting to POST them into the cluster.
* Added an integ test which asserts index templates are not overwritten by prior versions
* Multiple whitespace/formatting fixes from IDE auto-formatter. Suggest viewing this PR with whitespace changes hidden.

-----

### DIFF
--- a/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
+++ b/data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
@@ -35,7 +35,7 @@ public class DataPrepperConfiguration {
         try {
             return OBJECT_MAPPER.readValue(file, DataPrepperConfiguration.class);
         } catch (IOException e) {
-            throw new IllegalArgumentException("Invalid DataPrepper configuration file.");
+            throw new IllegalArgumentException("Invalid DataPrepper configuration file.", e);
         }
     }
 
--- a/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
+++ b/data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
@@ -17,6 +17,10 @@ import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.indices.CreateIndexRequest;
 import org.elasticsearch.client.indices.GetIndexRequest;
+import org.elasticsearch.client.indices.GetIndexTemplatesRequest;
+import org.elasticsearch.client.indices.GetIndexTemplatesResponse;
+import org.elasticsearch.client.indices.IndexTemplateMetadata;
+import org.elasticsearch.client.indices.IndexTemplatesExistRequest;
 import org.elasticsearch.client.indices.PutIndexTemplateRequest;
 import org.elasticsearch.common.unit.ByteSizeUnit;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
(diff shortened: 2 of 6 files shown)


wrijeff:
Manually ran integ tests here - https://github.com/wrijeff/simple-ingest-transformation-utility-pipeline/actions

data-prepper-core/src/main/java/com/amazon/dataprepper/parser/model/DataPrepperConfiguration.java
```diff
@@ -35,7 +35,7 @@ public static DataPrepperConfiguration fromFile(File file) {
         try {
             return OBJECT_MAPPER.readValue(file, DataPrepperConfiguration.class);
         } catch (IOException e) {
```
yadavcbala, line 37: Nit: final?

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -79,12 +83,13 @@ public void start() throws IOException {
     LOG.info("Starting Elasticsearch sink");
     restHighLevelClient = esSinkConfig.getConnectionConfiguration().createClient();
     final boolean isISMEnabled = IndexStateManagement.checkISMEnabled(restHighLevelClient);
-    final Optional<String> policyIdOptional = isISMEnabled? IndexStateManagement.checkAndCreatePolicy(restHighLevelClient, indexType) : Optional.empty();
+    final Optional<String> policyIdOptional = isISMEnabled ? IndexStateManagement.checkAndCreatePolicy(restHighLevelClient, indexType) :
```
yadavcbala, line 86: +1

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
```diff
@@ -158,16 +170,65 @@ private void createIndexTemplate(final boolean isISMEnabled, final String ismPol
...
+      }
+    } else {
+      LOG.info("Index template {} does not exist and should be created", indexTemplateName);
+      return true;
+    }
+  }
+
+  private Optional<IndexTemplateMetadata> getIndexTemplateMetadata(final String indexTemplateName) throws IOException {
+    final IndexTemplatesExistRequest existsRequest = new IndexTemplatesExistRequest(indexTemplateName);
+    final boolean exists = restHighLevelClient.indices().existsTemplate(existsRequest, RequestOptions.DEFAULT);
```
chenqi0805, line 209: Instead of making a separate request to check existence, can we check response.getIndexTemplates().size() == 0?

chenqi0805 (commented):
High level comment: the only scenario I can think of missing is when one uses different versions of data-prepper in parallel ingesting into the same Elasticsearch backend, there could be a race condition between older and newer version of template creation. But I think that use case should not be recommended or should just be avoided.

yadavcbala (approved):
LGTM

data-prepper-plugins/elasticsearch/src/main/java/com/amazon/dataprepper/plugins/sink/elasticsearch/ElasticsearchSink.java
wrijeff: IIRC I got an ES exception because of the 404 returned when the template does not exist. Figured the "correct" way would be to make a separate "exists" call here. Not especially concerned with the extra call as this only happens on startup.

chenqi0805: Sure. Not blocking

chenqi0805 (approved):

Issue #561, opened 2021-05-04
Update integ tests to use latest publicly available image of Data Prepper 


wrijeff:
Presently we have [integ tests](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/main/data-prepper-core/integrationTest.gradle#L153-L154) which builds Data Prepper and starts 2 Docker containers to test the peer_forwarder. A potential improvement would be to instead: 
1. Build the image as normal, start 1 container with it
2. Start the 2nd container by using the `:latest` version of the released Data Prepper image on dockerhub

This would continue to test the peer_forwarder plugin, but also ensure that old and new versions of Data Prepper are compatible with each other. This test doesn't have to be part of the existing peer_forwarder test, could be split off into a new compatibility test.

Also we might consider splitting integ tests out of the `core` package and into each plugins respective packages, would be more manageable as the project grows.

PR #609, opened 2021-05-17
Infrastructure: add compatibility e2e tests and separate github workflow

chenqi0805:
Signed-off-by: qchea <qchea@amazon.com>

*Issue #, if available:*
#561 

*Description of changes:*

-----
This PR
(1) adds raw-span e2e compatibility tests
(2) separate e2e tests workflow


.github/workflows/data-prepper-trace-analytics-raw-span-compatibility-e2e-tests.yml
```diff
@@ -26,6 +26,4 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
     - name: Run raw-span end-to-end tests with Gradle
```
wrijeff, line 28: nit - compatibility end-to-end tests?

data-prepper-core/integrationTest.gradle
```diff
@@ -169,7 +189,44 @@ task rawSpanEndToEndTest(type: Test) {
...
+    finalizedBy removeDataPrepperNetwork
+}
+
+task rawSpanCompatibilityEndToEndTest(type: Test) {
+    dependsOn build
+    dependsOn startOdfeDockerContainer
+    def createDataPrepper1Task = createDataPrepperDockerContainer(
+            "rawSpanDataPrepperFromBuild", "data-prepper1", 21890, 4900, "/app/${RAW_SPAN_PIPELINE_YAML}")
+    def createDataPrepper2Task = createDataPrepperDockerContainerFromPullImage(
+            "rawSpanDataPrepperFromPull", "data-prepper2", 21891, 4901, "src/integrationTest/resources/${RAW_SPAN_PIPELINE_NO_TRACE_GROUP_YAML}")
```
wrijeff, line 211: Shouldn't these be using the same configuration yaml?

Also would we expect the test to pass given that Prepper2 is missing both a TraceGroupPrepper and PeerForwarder? Or are all requests being routed to Prepper1?

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -45,6 +45,8 @@
     private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<Map<String, Object>>() {};
     private static final int DATA_PREPPER_PORT_1 = 21890;
     private static final int DATA_PREPPER_PORT_2 = 21891;
+    private static final int DATA_PREPPER_MAIN_PORT = 21890;
+    private static final int DATA_PREPPER_LATEST_PORT = 21891;
```
wrijeff, line 49: Might suggest clarifying these names a bit (is LATEST the most recent local build, or is it the "latest" tagged docker image?). 

Or does it even matter knowing which build is on port, just re-use PORT_1 and PORT_2?

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -135,6 +137,66 @@ public void testPipelineEndToEnd() throws InterruptedException {
         );
     }
 
+    @Test
+    public void testPipelineCompatibilityEndToEnd() throws InterruptedException {
```
wrijeff, line 141: Seeing as this is very similar to the existing `testPipelineEndToEnd()` test - was it copied with the intent of making changes to the 2 methods individually? Or can they be kept as a single test for now?

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
```diff
@@ -135,6 +137,66 @@ public void testPipelineEndToEnd() throws InterruptedException {
...
+        final ExportTraceServiceRequest exportTraceServiceRequestTrace2BatchWithRoot = getExportTraceServiceRequest(
+                getResourceSpansBatch(TEST_SPAN_SET_2_WITH_ROOT_SPAN)
+        );
+        final ExportTraceServiceRequest exportTraceServiceRequestTrace2BatchNoRoot = getExportTraceServiceRequest(
+                getResourceSpansBatch(TEST_SPAN_SET_2_WITHOUT_ROOT_SPAN)
+        );
+
+        sendExportTraceServiceRequestToSource(DATA_PREPPER_MAIN_PORT, exportTraceServiceRequestTrace1BatchWithRoot);
+        sendExportTraceServiceRequestToSource(DATA_PREPPER_MAIN_PORT, exportTraceServiceRequestTrace2BatchNoRoot);
+        sendExportTraceServiceRequestToSource(DATA_PREPPER_MAIN_PORT, exportTraceServiceRequestTrace1BatchNoRoot);
```
wrijeff, line 158: nit - I noticed the 3rd request is sent to `DATA_PREPPER_PORT_2` in the original test, any reason for it to differ here?

data-prepper-core/integrationTest.gradle
chenqi0805: As you noticed TraceGroupPrepper is missing and thus the separation of config yaml. The current testing logic is to send a portion of traceGroup with root span through the pulled data-prepper and then the data-prepper in the main branch should be able to pull root span from ES backend by the TraceGroupPrepper.

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
chenqi0805: > Or does it even matter knowing which build is on port, just re-use PORT_1 and PORT_2?

As mentioned above, for the current testing logic it is better to distinguish the data-prepper with trace-group prepper from the one w/o that prepper. But it might evolve.

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
chenqi0805: > was it copied with the intent of making changes to the 2 methods individually?
Somewhat. It is similar except the diff you noticed below.

data-prepper-core/src/integrationTest/java/com/amazon/dataprepper/integration/EndToEndRawSpanTest.java
chenqi0805: Yes, it is due to OTelTraceGroupPrepper missing in the latest pull image. Thus the whole traceGroup 1 is directed towards the current bleeding edge.

data-prepper-core/integrationTest.gradle
wrijeff: Ah, I forgot that the latest released image does not have that prepper. So `RAW_SPAN_PIPELINE_NO_TRACE_GROUP_YAML` exists to be compatible with the latest docker image, while `RAW_SPAN_PIPELINE_YAML` is compatible with the local build.

Could we instead have _both_ running containers use a config file which is compatible with the most recent release (i.e. both containers using PeerForwarder instead of one having a TraceGroupPrepper)? The intent would be to simulate a user upgrading their deployment without updating their config. 

The test should still pass if PeerForwarders are introduced. `RAW_SPAN_PIPELINE_NO_TRACE_GROUP_YAML` would be something more like `RAW_SPAN_PIPELINE_LATEST_RELEASE_COMPATIBLE` (name tbd).

chenqi0805: Sure. Sounds a good strategy.

data-prepper-core/integrationTest.gradle
```diff
@@ -169,7 +189,44 @@ task rawSpanEndToEndTest(type: Test) {
...
+    finalizedBy removeDataPrepperDockerContainer(stopDataPrepper1Task as DockerStopContainer)
+    finalizedBy removeDataPrepperDockerContainer(stopDataPrepper2Task as DockerStopContainer)
+    finalizedBy removeDataPrepperNetwork
+}
+
+task rawSpanCompatibilityEndToEndTest(type: Test) {
+    dependsOn build
+    dependsOn startOdfeDockerContainer
+    def createDataPrepper1Task = createDataPrepperDockerContainer(
+            "rawSpanDataPrepperFromBuild", "data-prepper1", 21890, 4900, "/app/${RAW_SPAN_PIPELINE_YAML}")
```
wrijeff, line 209: RAW_SPAN_PIPELINE_LATEST_RELEASE_YAML here too?

wrijeff (approved):
👍

data-prepper-core/integrationTest.gradle
chenqi0805: Thanks for the catch, missed that part.

PR #622, opened 2021-05-25
Fixed Gradle deprecation warnings.

dlvenable:
*Issue #, if available:* N/A

*Description of changes:*

I fixed some Gradle deprecation warnings. I also noticed that some projects are intended to be libraries, though they were not building that way. I change a few of them to get their transitive dependencies working with `api` dependencies.

I don't think we will be able to support Gradle 7 and JDK 16 yet because of dependencies on OpenSearch/ElasticSearch. Regardless, I hope these changes get us ready for such an update.

-----


wrijeff (approved):
👍

PR #671, opened 2021-06-16
Refactor of ServiceMapStatefulPrepper, moved MapDbPrepperState from f…

wrijeff:
…ile to heap.

Signed-off-by: Jeff Wright <74204404+wrijeff@users.noreply.github.com>

*Issue #, if available:*

*Description of changes:*
While attempting to benchmark a Data Prepper instance configured for only the service-map usecase, I found that it could not handle more than ~5 minutes of sustained load before locking up and eventually running out of memory. This PR introduce 3 fixes which enable the service-map prepper to run for at least an hour without any performance degradation.

#### Heap-based MapDB
Primary change here is moving from a file-based MapDB to heap-based. Basic benchmarks comparing file vs heap-based showed that file-based was an order of magnitude slower when inserting items. Simply switching from file to heap showed the biggest gain and allowed the instance to survive more than 5 minutes.

#### CyclicBarriers
Being able to run the test for longer revealed a race condition involving the current implementation's countdown latches ([code link](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/main/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java#L189-L196)). We expect for "non-master" threads to reach the `waitForRotationFinish();` before the master thread finishes `rotateWindows();` and calls `resetWorkState();`, however this isn't guaranteed. The race condition happens when the master thread resets the windowRotationLatch before other threads have a chance to await it. This results in a thread getting stuck forever. This change swaps out countdown latches for cyclic barriers to cut down on some of the synchronization complexity.

#### Re-use of window maps
It was observed that "old gen" heap space was ever growing, up to a point where garbage collection would kick in. This was caused by creating new `MapDbPrepperState` maps being created on window rotation ([code link](https://github.com/opendistro-for-elasticsearch/data-prepper/blob/main/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java#L340-L343)). As the windows are long-lived objects, they survive past the eden heap space and exist in old gen. Eventually they would accumulate to the point where a full GC was needed. This can be avoided by re-using long-lived objects instead of creating new ones.

-----

### DIFF
--- a/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
+++ b/data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
@@ -7,8 +7,8 @@ import com.amazon.dataprepper.model.configuration.PluginSetting;
 import com.amazon.dataprepper.model.prepper.AbstractPrepper;
 import com.amazon.dataprepper.model.record.Record;
 import com.amazon.dataprepper.plugins.prepper.state.MapDbPrepperState;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.Sets;
 import com.google.common.primitives.SignedBytes;
 import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
 import org.slf4j.Logger;
@@ -21,13 +21,11 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Set;
 import java.util.TreeMap;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.BrokenBarrierException;
+import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.BiFunction;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 @SingleThread
 @DataPrepperPlugin(name = "service_map_stateful", type = PluginType.PREPPER)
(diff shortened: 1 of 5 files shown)


data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -46,17 +44,16 @@
...
-    private static CountDownLatch windowRotationLatch = new CountDownLatch(1);
+    private static CyclicBarrier allThreadsCyclicBarrier;
+
     private volatile static MapDbPrepperState<ServiceMapStateData> previousWindow;
     private volatile static MapDbPrepperState<ServiceMapStateData> currentWindow;
     private volatile static MapDbPrepperState<String> previousTraceGroupWindow;
     private volatile static MapDbPrepperState<String> currentTraceGroupWindow;
     //TODO: Consider keeping this state in a db
-    private volatile static HashSet<ServiceMapRelationship> relationshipState = new HashSet<>();
+    private static final Set<ServiceMapRelationship> RELATIONSHIP_STATE = Sets.newConcurrentHashSet();
```
chenqi0805, line 54: 👍

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
```diff
@@ -165,88 +165,81 @@ private static File createPath(File path) {
...
+
+            serviceDependencyRecords.addAll(iteratePrepperState(previousWindow));
+            serviceDependencyRecords.addAll(iteratePrepperState(currentWindow));
+            LOG.info("Done evaluating service map edges");
+
+            // Wait for all workers before rotating windows
+            allThreadsCyclicBarrier.await();
 
             if (isMasterInstance()) {
                 rotateWindows();
```
chenqi0805, line 180: Is it possible to insert rotateWindows into the barrierAction? 
```
public CyclicBarrier(int parties, Runnable barrierAction)
```
Then we might get rid of the masterInstance and also avoid calling await twice

chenqi0805: I guess that might need removing the @SingleThread concurrency protocol as prerequisite. Feel free to push back to separate PR if feasible.

chenqi0805 (approved):
Just a question/suggestion. Nothing blocking. LGTM!

data-prepper-plugins/service-map-stateful/src/main/java/com/amazon/dataprepper/plugins/prepper/ServiceMapStatefulPrepper.java
wrijeff: Yep, will eventually phase out the concept of `masterInstance` and remove the `@SingleThread` annotation as you mentioned (in a separate PR though).

wrijeff: > Is it possible to insert rotateWindows into the barrierAction?

Interesting, hadn't see that 👍. Should be possible, though would have to research a bit more to confirm the order of events is what we expect.

JonahCalvo (commented):
👌

PR #733, opened 2021-07-19
Only register the Health gRPC service when it is configured to be used.

dlvenable:
Signed-off-by: David Venable <dlv@amazon.com>

*Issue #, if available:* 

None

*Description of changes:*

While reviewing the code, I noticed that the `OTelTraceSource` always registers a `HealthGrpcService`. We have a boolean flag which appears to be how we should determine whether to include it. I made this change to only conditionally add the service. Additionally, I added unit tests for both scenarios.

-----


wrijeff (commented):
Thanks for catching and fixing this 👍

wrijeff (approved):

chenqi0805 (approved):

